PR #64, opened 2013-02-03
Vector space reorganization, new macros and random forest example.

tixxit:
This is a bit diverse of a changeset. This includes:

Vector space type class reorganization. Most important, CoordinateSpaces now exist and we have instances for Seq-likes and Arrays. Also, InnerProductSpace extends VectorSpace now, not NormedVectorSpace. This gives us a dot product without requiring an NRoot. However, if an NRoot exists, then we can get an implicit instance of NormedVectorSpace (w/ (`x dot x).sqrt` as the norm).

There are 2 new macros. They are used to convert the RHS to a compatible type before applying an op. What this really means is that:

```
def twice[A: Ring](a: A) = a * 2
```

No longer boxes. The macro will call fromInt on the 2nd implicit (to *) first before applying *. It actually just calls from{TypeName} for whatever type is on the right.

A new Random Forest example exists now too. It can do both classification and regression and the example will cross-validate with 3 different data sets. It's quite nifty.

One note, I tried to specialize this example, but couldn't. The examples just wouldn't use the @spec'd variants. @non -- I'm sure I've missed something important. Could you take a look whenever you have some free time?

### DIFF
--- a/examples/src/main/scala/spire/example/DataSets.scala
+++ b/examples/src/main/scala/spire/example/DataSets.scala
@@ -4,6 +4,7 @@ import spire.algebra._
 import spire.math.Rational
 import spire.implicits._
 
+import scala.{ specialized => spec }
 import scala.annotation.tailrec
 import scala.collection.IterableLike
 import scala.collection.generic.CanBuildFrom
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spire/algebra/seq.scala
```diff
@@ -91,6 +91,24 @@ with InnerProductSpace[SA, A] {
...
+trait SeqCoordinateSpace[A, SA <: SeqLike[A, SA]] extends SeqInnerProductSpace[A, SA]
+with CoordinateSpace[SA, A] {
+  def coord(v: SA, i: Int): A = v(i)
+
+  override def dot(v: SA, w: SA): A = super[SeqInnerProductSpace].dot(v, w)
+
+  def axis(i: Int): SA = {
+    val b = cbf()
+
+    @tailrec def loop(j: Int): SA = if (i < dimensions) {
```
rtyley, line 103: For any valid `i` (`i < dimensions`), this condition is always true, so **recursion never terminates** - it's `j` that's getting incremented, not `i`!

```suggestion
    @tailrec def loop(j: Int): SA = if (j < dimensions) {
```

PR #159, opened 2013-10-09
Multivariate Polynomials

jamesthompson:
Added: 

Monomials - A class & companion object with associated Eq[Monomial] & Order[Monomial] traits
- monomials are a coefficient of type C (which must have a Ring[C] and Order[C] in scope) and its product with a list of variables (symbol given by a Char) to a power (Int). This is encoded as a Map[Char, Int].

Multivariate Polynomials - A class & companion object.
- multivariate polynomials are encoded by an Array[Monomial[C]]. A multivariate polynomial has the same algebraic properties as regular Polynomials. Although here they are not yet extending the Polynomial trait. The companion object has a value representing the Order[Monomial[C]] as lexicographic. There are also Glex and Grevlex Order[Monomial[C]] traits but these aren't yet being used.

MultivariatePolynomialTest - Tests for Monomial and MultivariatePolynomial with assertions on hard coded real solutions to arithmetic challenges. Also arbitrary term algebraic property tests for both Monomial and Multivariate Polynomial are tested. 

At present the easiest way to make a MultivariatePolynomial is through it's companion object (which must be used) as follows:
MultivariatePolynomial(Monomial(r"2/3", 'x' -> 2, 'y' -> 3, 'z' -> 1), Monomial(r"3/4", 'y' -> 2, 'z' -> 3), Monomial(r"4/5", x -> 0))


zmre:
Any chance this might get fixed up and merged in?

zmre:
Also, at a glance, it looks like this would have trouble representing something like `3x^2 + 2xy +x^2y^2`.  Is that true?  Or if this pull request were accepted, would that be possible?

jamesthompson:
@zmre This wouldn't have trouble representing the expression you mention above, I think the parser I wrote could even read this string and manipulate the resulting expression as is on this branch.

 I think the codebase has moved on a lot since I was last working on this. Perhaps @non can comment on whether he'd like to fork from this, start again or drop it. I'm not wedded to any particular outcome. I am really out of the loop with Spire at the moment and I'm not sure I could reasonably take this on right now :-/ 

If you wanted to take it and run with it, please feel free!!

tixxit:
@zmre While I'd love to have good multivariate polynomial support, I think someone else would likely need to do the work of fixing this branch up and maintaining it moving forward. Alternatively, it may be a good candidate for spire-extras or a 3rd party lib that depends on Spire too.

zmre:
@jamesthompson @tixxit I might be willing to take this on.  I looked closer at the implementation and it looks reasonable.  I might cherry pick some of the suggestions from the thread.  Are there any required changes that are missing right now that are prerequisites to accept an updated PR?

denisrosset:
Closed for now due to clean-up.

PR #166, opened 2013-10-31
Eigenvalue computation and more

luc-j-bourhis:
Guide to review

Most commits are short: the commit message and the code documentation should 
not require I expound on them here. I will only concentrate on the big picture 
and the milestone commits.
- cforRange macros: dd62010 and cb9d80d make extended use of them for the code 
  that shall be efficient.
- Synthetic matrix samples for testing: I started with a simple-minded one
  with 3df6735 but the important milestone came with 486250a and the generation
  of matrices with specific eigenvalue and conditioning properties. This is 
  essential to write proper tests such as those added in f5933fc for
  the Hessenberg decomposition, and those committed with the Schur decomposition
  in 2f23644.
- As a step toward the last goodie, I needed to generate "uniformly" distributed
  orthogonal matrices: 1905be4. The tests in that commit require a library that
  I had to install by hand, c.f. the comments. That means they will fail if you
  don't install it or change the build system.
- In ac90564, I remove "final case" from class Matrix because I wanted to 
  inherit from it to create test matrices. Probably not a good idea since,
  as I understand, you had added "final case" to make Matrix more efficient.
- WorkingArea was the biggest design failure of my first series of commits.
  It was not thread-safe at all and as a result, since the tests run in parallel,
  they were failing randomly. So I had to do a complet U-turn. So now I pass
  a scratchpad around but I make it an implicit argument so as to remove clutter.
  Commit db0b127 is where the big change came.
- I decided to fit Matrix with methods giving easy access to the best 
  implementation of eigen decomposition coded so far. So if I implement
  blocked Hessenberg decomposition or the multi-shift implicit QR algorithm, I
  plan to change the implementation of the method in Matrix to use that.
  So basically tools like Hessenberg.Decomposition, Schur.RealDecomposition, etc
  should only be used by advanced writers of client code, if at all. The commit
  is a1b37d6.


denisrosset:
Closing for now due to clean-up.

PR #304, opened 2014-08-14
Signed[A].abs is made optional, to be used when A is an additive group

denisrosset:
Hi all,

some mathematical objects are signed (or have parity), without necessarily having an absolute value.

For example, permutations have either sign +1 or -1, but the absolute value cannot be computed. There are several options:

i) leave `Signed` as is, and have users implement `def abs(a: A) = sys.error("Not implemented")` for such objects,
ii) write in the type class `def abs(a: A)(implicit ev: this.type <:< AdditiveGroup[A])` which has the advantage of being "free" especially when inlined
iii) write `def abs(a: A)(implicit ev: AdditiveGroup[A])` which requires the side-effect instantiation of `AdditiveGroup[A]`
iv) create an additional `SignedAbs` typeclass, which can be provided automatically from any `A: Signed: AdditiveGroup`, with specialized implementations for the primitive types

I tried to implement ii), but there is a conflict with `IsReal[A]` which does not inherit `AdditiveGroup[A]`. Having `IsReal` inherit `AdditiveGroup` causes implicit resolution to fail because of ambiguity when implicits `Field[A]` and `IsReal[A]` clash.

I dislike the solution iii) because `abs` can be called quite often in user code, on types like Double, etc..., where the actual computation is very fast and any overhead is painful.

The status quo is not good because `Signed` is probably going to be migrated to non/algebra at some stage, and it should not have this kind of idiosyncrasies.

### DIFF
--- a/core/src/main/scala/spire/algebra/Signed.scala
+++ b/core/src/main/scala/spire/algebra/Signed.scala
@@ -14,7 +14,8 @@ trait Signed[@spec(Double, Float, Int, Long) A] {
   def signum(a: A): Int
 
   /** An idempotent function that ensures an object has a non-negative sign. */
-  def abs(a: A): A
+  def abs(a: A)(implicit ev: AdditiveGroup[A]): A =
+    if (signum(a) < 0) ev.negate(a) else a
 
   def isZero(a: A): Boolean = signum(a) == 0
 }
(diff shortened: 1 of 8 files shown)


denisrosset:
The typeclass "Signed" should be improved, see for example https://en.wikipedia.org/wiki/Linearly_ordered_group for a mathematical definition.

But `abs` should definitively not be optional. I'm changing my mind and closing this.

PR #311, opened 2014-10-19
Partial move of abs from Signed to IsReal. Does not compile.

denisrosset:
Continued from #304.

I tried to move .abs to IsReal, the problem is that IsReal is not applied everywhere: for example, the Scala Numeric wrappers do not use IsReal, the vector space norm implementations have to be converted, and Complex does not implement IsReal either.

This is the WIP.


denisrosset:
No longer relevant.

PR #341, opened 2014-12-15
Partial algebras.

denisrosset:
Split Action into LeftAction and RightAction.
Renamed the Action syntax classes accordingly instead of GroupOps, PointOps.
Added PartialAction implementation, with both Option-like and PartialFunction-like semantics.
Added implementation of Semigroupoid, PartialMonoid, Groupoid.
Added syntax for partial operations, with both Option-like and PartialFunction-like semantics.


denisrosset:
@non : I see that a lot of work is being done on non/algebra, I will have a look at it.

This can wait until the work there is finished.

non:
Yeah sorry I haven't been more responsive. I'm definitely giving it some thought.

non:
One note to add: if we are willing to not require `Group[A]` to inherit `Groupoid[A]`, but to support an implicit wrapper instead, then it will be very easy to add these to Spire first and wait and see what happens in *algebra*, since we won't be required to have them there.

What do you think about this? @tixxit?

denisrosset:
We could also use inheritance the other way around, e.g. have partial algebras inherit the regular ones --- this is actually what is done in `PartialFunction[A, B] extends (A => B)`.

We would thus remove the Nullbox / non-Nullbox distinction and use Nullbox everywhere.

There would be an implicit wrapper for the regular algebras used partially, so a small overhead --- except if we can write a macro that generates a `.isInstanceOf[Partial...]` stub.

I am waiting for @tixxit before making a move.

non:
I would prefer to avoid inheriting in that direction -- the `PartialFunction` vs `Function` relationship has been causing a lot of pain for Scala users for some time now. Also, I would rather not use `Nullbox[A]` except in cases where we really need it, since in some sense it "de-specializes" primitives.

denisrosset:
Note: this is superseded by #353, #356 and #357.

Please have a look at #356 first.

PR #352, opened 2015-01-28
Partial algebras, second try

denisrosset:
This is a reimplementation of #341, with the following changes:

- the partial algebra hierarchy will be separated from the standard hierarchy
- the partial algebra hierarchy will be implemented using Nullbox
- implicit conversions will be provided from regular to partial algebras
- the Nullbox extractor has been adapted for Scala 2.10 (at the price of `Option` instantiations for 2.10 only)


denisrosset:
Sorry, using bad root for the PR. I am redoing it.

PR #353, opened 2015-01-28
Partial algebras, second try

denisrosset:
This starts a reimplementation of #341, with the following choices:

- the partial algebra hierarchy will be separated from the standard hierarchy
- the partial algebra hierarchy will be implemented using Nullbox
- implicit conversions will be provided from regular to partial algebras
- the Nullbox extractor has been adapted for Scala 2.10 (at the price of `Option` instantiations for 2.10 only)


denisrosset:
@non : there is still one problem. The partial and regular algebras share two methods: `.isId` and `.inverse`; so they have to use the same enrichment Ops class to provide these methods.

There are few options here:

- either have both `Group` and `Groupoid` inherit traits that provide the methods needed (this is the option I choose here, with two traits `HasIsId` and `HasInverse`), but they would have to be provided on non/algebra as well,
- either have a trick in the implicit conversion to `XXXOps` class, for example using [union types](http://www.chuusai.com/2011/06/09/scala-union-types-curry-howard/)

What do you think ?

denisrosset:
Also pinging @tixxit

denisrosset:
Arrgg, Scala 2.10 needs its own hashCode and equals.

denisrosset:
Note: the HasIsId and HasInverse split is needed because otherwise, the Group -> Groupoid conversion is kicking in when calling (a: Group[A]).inverse, and then there are problems ambiguous implicits. Another advantage is having a single typeclass per enriched method.


There are two other solutions:

- making the Group -> Groupoid implicit conversion low-priority, but how robust is it ?
- call the isId and inverse methods partialIsId and partialInverse, even if they always suceed (by the algebraic laws)

denisrosset:
2nd note: I don't have a PartialMonoid instance in this PR, because I did not find a use case for it either in my own work or in the mathematical litterature.

Thus, the partial algebras are Semigroupoid and Groupoid.

denisrosset:
Completed partial algebraic structures.

tests/src/test/scala/spire/PartialSyntaxTest.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  implicit val IntGroup: Group[Int] = implicitly[AdditiveGroup[Int]].additive
+  implicit val SeqIntEq: Eq[Seq[Int]] = spire.optional.genericEq.generic[Seq[Int]]
+
+  test("Semigroupoid syntax")(check(forAll { (a: Seq[Int], b: Seq[Int]) => testSemigroupoidSyntax(a, b) }))
+  test("Groupoid syntax")(check(forAll { (a: Seq[Int], b: Seq[Int]) => testGroupoidSyntax(a, b) }))
+
+  def testSemigroupoidSyntax[A: Semigroupoid: Eq](a: A, b: A) = {
+    import spire.syntax.semigroupoid._
+    ((a |+|? b) == Semigroupoid[A].partialOp(a, b)) &&
```
non: Since we are using `Eq` here we should probably be using `===` instead of `==` to test equality.

non:
Hey this looks great! I think there are just three issues in this PR:

1. We should probably come up with a better name than `Nullbox`. I just slapped that name on the idea and I worry that it is not very clear. Any suggestions? `Opt[_]`?

2. It is possible we will later need to remove `HasInverse` and `HasIsId` and just duplicate the Ops classes. Once `algebra` is published and we start porting Spire to depend on it we'll be constrained by its structure. I don't think this issue blocks the PR but just wanted to let you know.

3. There was a minor issue with using `==` instead of `===` in the tests that I commented on.

I could also imagine using ScalaCheck to test `Nullbox[_]` as well but honestly the tests right now seem good, so I'm happy to worry about that later and merge this sooner, once we resolve (1) and (3).

What do you think @tixxit ?

denisrosset:
1. Agree with `Nullbox[_]` becoming `Opt[_]`; I did a quick Google search, the name is already used in a book example for the exact same class, an option-like derivative wrapping a reference. Waiting for @tixxit approval.
2. I will have a try without, by demoting the Group -> Groupoid implicit conversion to a lower priority using traits.
3. For a while, I did not have a `Eq[Nullbox[_]]` instance; thanks for spotting the issue!

tixxit:
`Opt[_]` is good.

non:
Looks great! I'm merging now. Thanks for all your hard work navigating the murky waters of this PR, @denisrosset.

PR #356, opened 2015-01-29
Action split into LeftAction and RightAction

denisrosset:
Hi all,

here is a split of Action into LeftAction and RightAction, in case the underlying transformation cannot be inverted.

Of course, provided the acting element comes from a group, the left action defines a right action and vice-versa.

Note: this is a prologue to my implementation of partial actions, where this split is needed. But I prefer to provide PRs in tiny pieces.

This single PR can be merged without too much discussion.


non:
Looks good to me. :+1: What do you think @tixxit ?

PR #357, opened 2015-01-29
Partial actions

denisrosset:
# DO NOT MERGE

Replaced by #360.

Note: ~~work in progress~~ depends on both ~~#353~~ (since merged) and #356, which should be merged first.


denisrosset:
At feature parity with outdated #341.

PR #359, opened 2015-01-29
Improved compatibility with future non/algebra traits.

denisrosset:
Follow-up of #353 

I found a trick to avoid ambiguous implicits between the regular and partial hierarchies, at the price of boxing the `Groupoid[A].inverse` and `.isId()` method calls (for now).

Correcting also Action syntax that was lost in the merge.


denisrosset:
Wow, either I'm messing up my trees, or GitHub does not like so much overlapping PRs.

Can you please merge this correction as soon as possible, and I will restart #357 from scratch ?

non:
Looks great, thanks!

PR #360, opened 2015-01-30
Partial actions

denisrosset:
Implementation of partial actions, i.e. left or right actions that can either succeed or fail, with the result represented as an `Opt[A]`.

An optional implementation of permutations using `Map[Int, Int]` is provided in `spire.optional`, to provide the material for syntax and law tests. This part could be moved completely in the tests package (what do you think, @non and @tixxit ?); a complete implementation of permutations, finite groups, etc..., is outside the scope of Spire.

Supercedes and replaces #357.

### DIFF
--- a/core/src/main/scala/spire/algebra/partial/PartialAction.scala
+++ b/core/src/main/scala/spire/algebra/partial/PartialAction.scala
@@ -19,7 +19,7 @@ import spire.util.Opt
   * `(g.rightId ?|+|> p).get === p`, the operation `?|+|>` being defined.
   */
 trait LeftPartialAction[P, G] extends Any {
-  def actlIsDefined(g: G, p: P): Boolean
+  def actlIsDefined(g: G, p: P): Boolean = partialActl(g, p).nonEmpty
   def partialActl(g: G, p: P): Opt[P]
 }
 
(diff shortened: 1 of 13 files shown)


denisrosset:
@non, after we settle this, may I ask you for a release, or do you prefer to wait on non/algebra ? I'd like to release my library Alasc to Sonatype...

core/src/main/scala/spire/algebra/partial/PartialAction.scala
```diff
@@ -0,0 +1,80 @@
...
+  @inline final def apply[P, G](implicit G: RightPartialAction[P, G]) = G
+
+  implicit def fromRightAction[P, G](implicit G: RightAction[P, G]): RightPartialAction[P, G] =
+    new RightPartialAction[P, G] {
+      override def actrIsDefined(p: P, g: G) = true
+      def partialActr(p: P, g: G): Opt[P] = Opt(G.actr(p, g))
+    }
+}
+
+trait PartialAction[P, G] extends Any
```
tixxit: It would be nice if there were some scaladocs here, even if they just referenced `{Left,Right}PartialAction`.

tixxit:
We should probably add the new type classes to the README too.

tixxit:
Aside from those 2 issues, LGTM.

denisrosset:
Thanks @tixxit.

Will perform the ScalaDoc update and update the README for the regular `Action` as well.

PR #361, opened 2015-01-31
Topic/interval ops etc

non:
Hey, these are some nice changes I made to intervals preparing this talk.

(I don't think any of this should be very controversial.)

Review by @tixxit and/or @denisrosset.


core/src/main/scala/spire/math/Interval.scala
```diff
@@ -421,6 +423,9 @@ sealed abstract class Interval[A](implicit order: Order[A]) { lhs =>
       case (_, _) => lhs * rhs.reciprocal
     }
 
+  def /(rhs: A)(implicit ev: Field[A]): Interval[A] =
+    lhs * rhs.reciprocal
```
tixxit: Why use this rather than `lhs / rhs`? I'm worried a bit about using 2 lossy operations for, eg, Double, rather than just 1.

core/src/main/scala/spire/math/Interval.scala
```diff
@@ -706,6 +738,31 @@ object Interval {
 
   def apply[A: Order](lower: A, upper: A): Interval[A] = closed(lower, upper)
 
+  /**
+   * Return an Interval[Rational] that corresponds to the error bounds
+   * for the given Double value.
+   * 
+   * The error bounds are represented as a closed interval, whose
```
tixxit: We should probably update the docs to mention the positive/negative infinity case.

core/src/main/scala/spire/math/Interval.scala
non: I could have written `lhs / Interval.point(rhs)` but if you read that method's body I've just inlined it here (without the `point` call).

We don't currently have a *direction division* implementation for `Interval[_]` (the last one I wrote was a bit buggy and using reciprocal was easier to prove correct).

core/src/main/scala/spire/math/Interval.scala
non: Oh yeah, that was a big oversight. Done!

core/src/main/scala/spire/math/Interval.scala
tixxit: Ah, I see. Sorry for the noise!

tixxit:
:+1:

core/src/main/scala/spire/math/Interval.scala
non: No worries, it's a good reminder that maybe I should take another crack at doing it right.

PR #367, opened 2015-02-04
Added override to LongRational.equals

### LINKED ISSUE #366: Rational.equals very slow for LongRational
Comparing two LongRational using equals/==/=== is much slower than it has to be. 

Here is a little benchmark that compares == to a "fastEquals" method

    Rational primitive operations benchmark Integer 12345 67890
    Benchmark for a == b (20 calls in 311.3 ms)
      Time:    58.94 ns   95% CI 58.29 ns - 59.58 ns   (n=20)
      Garbage: 0.1907 ns   (n=1 sweeps measured)
    Benchmark for a fastEquals b (20 calls in 495.0 ms)
      Time:    5.889 ns   95% CI 5.852 ns - 5.925 ns   (n=20)

And here is the "fastEquals" method:

    def fastEquals(a:Rational, b:Rational) = (a, b) match {
      case (LongRational(an,ad), LongRational(bn, bd)) ⇒
        an == bn && ad == bd
      case _ ⇒
        a == b
    }

I think the current equals method is less than optimal since it uses RationalLike

    override def equals(that: Any): Boolean = that match {
      case that: Real => this == that.toRational
      case that: Algebraic => that == this
      case that: RationalLike => num == that.num && den == that.den // <= this is boxing
      ...
    }

Overriding equals for LongRational to check rhs for LongRational should fix this.

rklaehn:
ensures that a comparison between two LongRational numbers happens
without any boxing. Speeds up == between two LongRationals
significantly.

Fixes #366

### DIFF
--- a/core/src/main/scala/spire/math/Rational.scala
+++ b/core/src/main/scala/spire/math/Rational.scala
@@ -763,6 +763,11 @@ private[math] object LongRationals extends Rationals[Long] {
         else
           (SafeLong(n) * (r.d / dgcd) - SafeLong(r.n) * (d / dgcd)).signum
     }
+
+    override def equals(that: Any) = that match {
+      case that: LongRational => this.n == that.n && this.d == that.d
+      case _ => super.equals(that)
+    }
   }
 }
 


non:
This looks great.

The Travis error seems transient. I'm re-running the job and will sign-off when it's done.

non:
:+1:

denisrosset:
:+1:

PR #369, opened 2015-02-05
Override hashCode separately in LongRational and BigRational

### LINKED ISSUE #368: Rational.hashCode is very slow for small, whole numbers
It seems that Rational.hashCode is less than optimal for small, whole numbers. Here is a little benchmark:

    import ichi.bench.Thyme
    import spire.math.Rational

    object RationalHashCodeBench extends App {
      val a = Rational(1,3)
      val b = Rational(3)
      val th = Thyme.warmed(warmth = Thyme.HowWarm.Bench, verbose = println)
      th.pbenchWarm(th.Warm(a.hashCode), title = s"$a.hashCode")
      th.pbenchWarm(th.Warm(b.hashCode), title = s"$b.hashCode")
    }

And the result:

    Benchmark for (1/3).hashCode (60 calls in 1.897 s)
      Time:    15.00 ns   95% CI 14.46 ns - 15.54 ns   (n=20)
      Garbage: 0.03576 ns   (n=6 sweeps measured)
    Benchmark for (3).hashCode (60 calls in 1.469 s)
      Time:    183.8 ns   95% CI 175.3 ns - 192.3 ns   (n=20)
      Garbage: 1.526 ns   (n=19 sweeps measured)

As you can see, calling hashCode on a whole number <= Long.MaxValue is *way* slower than it has to be. The reason is that the hashCode method checks if the number is small enough for the unifiedPrimitiveHashcode by converting it to a BigInt, among other things:

      override def hashCode: Int =
        if (isWhole && toBigInt == toLong) unifiedPrimitiveHashcode // toBigInt is wasteful for this check!
        else 29 * (37 * num.## + den.##)

I think a simple override of LongRational.hashCode should fix this. Whole numbers within the range of Long are always represented as LongRational, right? Or is it possible to have e.g. a BigRational(3,1)?

rklaehn:
In LongRational, we know that the numerator is within Long range, so we
always use unifiedPrimitiveHashcode if the denominator is 1. We also have
to override longValue, since that is used by unifiedPrimitiveHashcode.

In BigRational, we know that the numerator is larger than a long if the
denominator is 1, so we never have to use unifiedPrimitiveHashcode

Fixes #368

### DIFF
--- a/core/src/main/scala/spire/math/Rational.scala
+++ b/core/src/main/scala/spire/math/Rational.scala
@@ -441,7 +441,7 @@ private[math] abstract class Rationals[@specialized(Long) A](implicit integral: 
     def toBigInt: BigInt = (integral.toBigInt(num) / integral.toBigInt(den))
     def toBigDecimal: BigDecimal = integral.toBigDecimal(num) / integral.toBigDecimal(den)
 
-    def longValue = toBigInt.longValue    // Override if possible.
+    def longValue = toBigInt.longValue
     def intValue = longValue.intValue
 
     def floatValue = doubleValue.toFloat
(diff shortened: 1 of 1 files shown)


non:
:+1: assuming travis passes

thanks!

Issue #372, opened 2015-02-07
Interval multiplication: bug report provided by @rklaehn


denisrosset:
```scala
object RationalTest3 extends App {
  val a = Interval(-1, 1)
  val b = Interval.above(1)
  val x = -1
  val y = 10
  println(s"$a * $b = ${a*b}")
  println(a.contains(x))
  println(b.contains(y))
  println((a*b).contains(x*y))
}
```

denisrosset:
Actually, there are randomized tests in IntervalTest.scala, and the case above should have been exercised. I'm investigating why the current tests fail to verify it.

denisrosset:
Ok, found it. The `Interval` arbitrary generator is using a `Double` to generate different interval types, thinking the `Double` is between 0 and 1. It is not, so the test cases are heavily biased towards bounded intervals.

denisrosset:
Solved by #373 .

PR #373, opened 2015-02-07
Interval product bug

denisrosset:
Investigation of #372


non:
Alright, it seems like the invalid interval generator meant our sampling tests weren't uncovering as many issues as they could have.

Troubling, but better to discover now than later.

I'm going to branch your branch and see if I can get the tests reliably passing. I may need to do some of the restructuring you alluded to in Gitter.

denisrosset:
@non: all the subcases are explicitly covered now. I launched the randomized several times on my computer, everything seems fine. I'll let you have a look (especially the flag conversion details!!).

non:
Looks great to me. I'm merging now.

PR #383, opened 2015-02-17
Fix bug where fact(0) was returning 0 instead of 1.

### LINKED ISSUE #382: spire.math.fact(0) == 0
shouldn't this evaluate to 1 instead of 0?

non:
This fixes #382.


tixxit:
We should probably add a test for this too.

non:
Yes, that's a great point.

non:
(That previous test failure was the intermittent *float gcd* error, by the way.)

non:
Alright tests added and passed. Thanks @jinmago for catching this!

PR #386, opened 2015-02-18
Explicitly implement isValidX

### LINKED ISSUE #385: Bug in rational/int equality
There seems to be a bug in rational/int equality. Probably something going wrong with the "cooperative equality" code. I tested this with spire 0.8.2 to make sure it was not recently introduced:

```scala
scala>  Rational("18446744073709551614") == -2
res1: Boolean = true

scala>  Rational("18446744073709551614").isValidInt
res2: Boolean = true
```

The given number is Rational(Long.MaxValue) * 2. In case that helps to track it down.

rklaehn:
The implementation in the base class will not work for rationals.

x.toLong == x.toInt does not guarantee that the value is within int range
for values that can become larger than a long.

Fixes #385

However, there should be some tests to make sure this does not happen again. Also, SafeLong seems to suffer from similar issues which will have to be tracked separately.

### DIFF
--- a/core/src/main/scala/spire/math/Rational.scala
+++ b/core/src/main/scala/spire/math/Rational.scala
@@ -559,6 +559,16 @@ private[math] object LongRationals extends Rationals[Long] {
 
     override def signum: Int = java.lang.Long.signum(n)
 
+    override def isWhole: Boolean = d == 1L
+
+    override def isValidChar: Boolean = isWhole && n.isValidChar
+
+    override def isValidByte: Boolean = isWhole && n.isValidByte
+
+    override def isValidShort: Boolean = isWhole && n.isValidShort
+
+    override def isValidInt: Boolean = isWhole && n.isValidInt
+
     override def unary_-(): Rational =
       if (n == Long.MinValue) BigRational(-BigInt(Long.MinValue), BigInt(d))
       else LongRational(-n, d)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spire/math/Rational.scala
```diff
@@ -559,6 +559,16 @@ private[math] object LongRationals extends Rationals[Long] {
 
     override def signum: Int = java.lang.Long.signum(n)
 
+    override def isWhole: Boolean = n == 1L
```
non: Shouldn't this be `d == 1L`?

core/src/main/scala/spire/math/Rational.scala
```diff
@@ -879,6 +889,14 @@ private[math] object BigRationals extends Rationals[BigInt] {
 
     override def signum: Int = n.signum
 
+    override def isValidChar = false
```
non: Wouldn't you need to override `isWhole` in this case as well? Very large integers should still be whole numbers.

core/src/main/scala/spire/math/Rational.scala
rklaehn: You are right, of course.

core/src/main/scala/spire/math/Rational.scala
rklaehn: For BigRational I keep the isWhole implementation from RationalLike for now.

non:
Looks great. Thanks!

PR #389, opened 2015-02-18
A number of small improvements

rklaehn:
- The signum method is implemented in both LongRational and BigRational, so the base implementations in Rational and RationalLike can never be called

- Using numerator.signum to determine the sign is more efficient in than doing a comparison to Rational.zero

- Using UnsafeBuild for a denominator of 1 is unnecessary. We also don't need to figure out which rational variant will be used: a SafeLongBigInt will always produce a BigRational, and a SafeLongLong will always produce a LongRational. Likewise, Int and Long values will always produce a LongRational.

### DIFF
--- a/core/src/main/scala/spire/math/Rational.scala
+++ b/core/src/main/scala/spire/math/Rational.scala
@@ -354,8 +354,8 @@ object Rational extends RationalInstances {
 
   implicit def apply(n: SafeLong): Rational =
     n match {
-      case SafeLongLong(x) => LongRationals.unsafeBuild(x, 1L)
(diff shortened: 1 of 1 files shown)


non:
This looks great. Waiting on Travis -- will merge if the tests pass.

Especially good point about using `signum` instead of comparing to zero. I know we take advantage of that elsewhere, I guess somehow this slipped through the cracks.

rklaehn:
The advice of using signum instead of doing a full comparision is one of the things that should go into the Rational.md "design document" file.

non:
That Travis failure was just a resolution issue, so it's fine (man those Travis errors are annoying). Merging now.

PR #394, opened 2015-02-22
Added some tests to increase coverage of Rational

rklaehn:
Added some tests to increase coverage of Rational, and removed dead code that can not be tested because it is never called.

Also excluded the large piece of code for Rational.pow and Rational.nroot from coverage, because it is going to be removed soon anyway.

### DIFF
--- a/core/src/main/scala/spire/math/Rational.scala
+++ b/core/src/main/scala/spire/math/Rational.scala
@@ -18,6 +18,8 @@ sealed abstract class Rational extends ScalaNumber with ScalaNumericConversions 
   def numeratorAsLong: Long
   def denominatorAsLong: Long
 
+  def isValidLong: Boolean
+
   def isWhole: Boolean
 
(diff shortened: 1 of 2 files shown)


non:
You mean `Rational#fpow` right? I think `Rational#pow` with an integer argument is definitely staying.

rklaehn:
I mean this one:
```
def pow(exp: Rational)(implicit ctxt: ApproximationContext[Rational]): Rational
```

non:
Oh yeah, that one will be going away.

rklaehn:
By the way: the last commit is a fix for a completely trivial bug that I only discovered by mindlessly entering tests to get the coverage up. `Rational.apply(Rational.one: Number)` would result in infinite recursion. I think this underscores the need to increase test coverage to preferrably 100% even for seemingly trivial code.

The build failure is the float gcd thing.

non:
Agreed. I'll merge this now if you are happy with it.

By the way, I'm close to 100% on `SafeLong`.

PR #395, opened 2015-02-23
Bug/safelong coverage

non:
This PR introduces near 100% test covearge for
SafeLong, and fixes several issues that were
discovered while adding tests.

At this point the behavior of SafeLong and BigInt
should be identical for all values (except that
SafeLong will be faster).


non:
Review by @rklaehn, @tixxit.

non:
Yeah, you raise a good point about using `copy`. I'll move the types to be private.

rklaehn:
LGTM!

PR #396, opened 2015-02-24
Topic/vectorspace redux

pleira:
Update to current Spire's master. 

Have a look at IndexBasis as I commented out the size method. Also check LawTests. The RandomForest and DataSets examples are comented out as well.


denisrosset:
We are cleaning up the PR backlog, moving to an issue.

PR #398, opened 2015-02-27
Initial attempt of adding some documentation of the invariants of SafeLong and Rational

rklaehn:
@non I used a single md file at top level as discussed on gitter.


non:
I was on vacation without internet most of last week, so I'm just coming back to this now. I'll review it tonight or tomorrow.

non:
Just read through. This is great! Thanks!

Merging now.

PR #400, opened 2015-03-06
Implement a PCG-based Generator (fixes #399)

### LINKED ISSUE #399: PCG generator
I'd be happy to contribute a [PCG](http://www.pcg-random.org/)-based generator if it would be useful.

alexeyr:


non:
Thanks!

One question: are you using any comparisons or other operations where `ULong` behaves differently from `Long`? If not, I think it would be better to just use `Long` instead. Value classes are quite efficient but there are situations where primitives will still perform better.

alexeyr:
Yes, it should work the same with `Long` (validated by my Java implementation). Would you prefer I amend this commit or add a new one?

non:
I think you can just push an additional commit. Thanks.

PR #407, opened 2015-03-18
Fix bug in rational cooperative equality

### LINKED ISSUE #406: Error in cooperative equality of Rational and Long
```scala
scala> spire.math.Rational(1,2) == 0L
res3: Boolean = true
```

I think the reason for this is that equals calls into unifiedPrimitiveEquals, and unifiedPrimitiveEquals calls toLong without checking if it is even a whole number. It is mentioned in unifiedPrimitiveEquals docs that the method should only be called for types that are "happy to be converted to Long"... So many usages of this method in spire seem wrong.

In case of rational, calling isValidLong before calling unifiedPrimitiveEquals should fix the issue.

```
  /** Should only be called after all known non-primitive
   *  types have been excluded.  This method won't dispatch
   *  anywhere else after checking against the primitives
   *  to avoid infinite recursion between equals and this on
   *  unknown "Number" variants.
   *
   *  Additionally, this should only be called if the numeric
   *  type is happy to be converted to Long, Float, and Double.
   *  If for instance a BigInt much larger than the Long range is 
   *  sent here, it will claim equality with whatever Long is left
   *  in its lower 64 bits.  Or a BigDecimal with more precision
   *  than Double can hold: same thing.  There's no way given the
   *  interface available here to prevent this error.
   */
  protected def unifiedPrimitiveEquals(x: Any) = x match {
    case x: Char    => isValidChar && (toInt == x.toInt)
    case x: Byte    => isValidByte && (toByte == x)
    case x: Short   => isValidShort && (toShort == x)
    case x: Int     => isValidInt && (toInt == x)
    case x: Long    => toLong == x
    case x: Float   => toFloat == x
    case x: Double  => toDouble == x
    case _          => false
  }
```

rklaehn:
unifiedPrimitiveEquals is only supposed to be called for types that "are
happy to be converted to a Long", so we need to sort out non-long rationals
before.

Fixes #406

### DIFF
--- a/core/src/main/scala/spire/math/Rational.scala
+++ b/core/src/main/scala/spire/math/Rational.scala
@@ -491,7 +491,7 @@ private[math] abstract class Rationals[@specialized(Long) A](implicit integral: 
       case that: Natural => isWhole && this == Rational(that.toBigInt)
       case that: Complex[_] => that == this
       case that: Quaternion[_] => that == this
-      case that => unifiedPrimitiveEquals(that)
+      case that => isValidLong && unifiedPrimitiveEquals(that)
     }
 
     override def toString: String = if (den == 1L)


non:
These test failures seem unrelated to the bug fix. I'll investigate and merge this if possible.

non:
I think this fix is not right actually. See #408 for my proposed solution.

non:
(The reason is that `unifiedPrimitiveEquals` also supports types like `Double`.)

rklaehn:
You are right. I forgot about doubles.

PR #411, opened 2015-03-20
Unify and fix Polynomial's equals and hashCode methods.

### LINKED ISSUE #410: hashCode for polynomial does not work
```scala
scala> Polynomial("(1)").##
res1: Int = -135076863

scala> Polynomial("(1)").##
res2: Int = 776699639

scala> Polynomial("(1)").##
res3: Int = 797295762

scala> Polynomial("(1)").##
res4: Int = 639834089
```

Both PolyDense and PolySparse use arrays internally. So the hashCode method generated for case classes is not sufficient. Also, hashCode would have to be implemented in such a way that both PolyDense and PolySparse representations of the same polynomial return the same hash code.

At the very least, hashCode needs to be implemented to be 0 to fulfill the equals/hashCode contract.

non:
This commit adds a termsIterator method to Polynomial. This method
returns an Iterator[Term[C]] for all non-zero coefficients in the
polynomial, from smallest-exponent to largest. This iterator is
used to implement equality and hashCode in a standardized way that
works for dense and sparse polynomials.

The hashing formula used is essentially:

  val it: Iterator[Term[C]] = p.termsIterator
  val ns: Iterator[Int] = it.map(t => (0xfeed1257 * t.exp) ^ t.coeff.##)
  val hashCode: Int = ns.foldLeft(0)(_ ^ _)

One useful side-effect is that if the terms iterator is empty
(i.e. the polynomial is zero) the hash code is 0. Also, if p
consists of a single term whose exponent is zero, the overall
hashCode will be identical to the hash code of the term's
cofficient.

One gotcha with the termsIterator method is that it uses a
(_ == 0) test to filter out terms whose coefficient is zero.
This necessary in order to support universal equality, and is
documented in the method's ScalaDoc comment. We could choose
to make this method private to Spire if it is deemed to be
"too dangerous".

Fixes #410.

Review by @tixxit and @rklaehn.

### DIFF
--- a/core/src/main/scala/spire/math/Polynomial.scala
+++ b/core/src/main/scala/spire/math/Polynomial.scala
@@ -175,6 +175,17 @@ trait Polynomial[@spec(Double) C] { lhs =>
     lb.result()
   }
 
+  /**
+   * Return an iterator of non-zero terms.
+   *
+   * This method is used to implement equals and hashCode.
+   *
+   * NOTE: This method uses a (_ == 0) test to prune zero values. This
+   * makes sense in a context where Semiring[C] and Eq[C] are
+   * unavailable, but not other places.
+   */
+  def termsIterator: Iterator[Term[C]]
+
   /** Returns a map from exponent to coefficient of this polynomial. */
   def data(implicit ring: Semiring[C], eq: Eq[C]): Map[Int, C] = {
     val bldr = new scala.collection.mutable.MapBuilder[Int, C, Map[Int, C]](Map.empty[Int, C])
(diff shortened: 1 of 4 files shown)


rklaehn:
I guess this will work for spire types and scala primitives.

Polynomial.zero will not be equal to Polynomial.constant(zero), but I guess that is unavoidable.

The x == 0 thing will not work for types that don't attempt universal equality. So equality and hashCode will be broken for Polynomial[T] if T does not support checking for zero using == 0. Is that acceptable?

non:
I think `Polynomial.constant(zero)` will equal `Polynomial.zero` (assuming `zero == 0`), since the `termsIterator` ignores zero coefficients.

Yes, I think that if the underlying type doesn't support universal equality (which would require `zero == 0`) then universal equality won't work. So I think it's an acceptable trade-off.

tixxit:
@rklaehn I think the answer is that it is the best we can do, which kind of sucks... The only other thing we could do is capture a type class / zero for `DensePolynomial`. As usual, the real ideal is to push people to use type class based equality (`Eq`), and hashes once we have `algebra` integration.

rklaehn:
@non Yes, you are right. And Polynomial.zero.hashCode will be 0, which also is the "universal" hashCode of "zero" elements. I still need to get the hang of all this madness...

Anyway, I only need Polynomials of Rationals and maybe Doubles at work. So for my real use case at work, the solution will work.

non:
I'm just going to merge this -- I think at this point it's received enough review.

PR #412, opened 2015-03-22
Use unicode superscripts for exponents of polynomials

rklaehn:
Also allow unicode superscripts when parsing polynomials. The ascii version `x^2` is still supported.

- I put the unicode support code in poly.Term and made it private. Is there a better place to put it?

- I made a small additional change to parse. Before, `poly"(x² + x² + 1)"` would give `(x² + 1)` (everything but the last coefficient for each exponent is ignored). Now the terms are added, so `poly"(x² + x² + 1)"` gives `(2x² + 1)`. The downside is that parsing is slightly slower for large polynomials, but I think correctness trumps speed.


non:
The correctness fix is great, thanks!

:+1: 

I'm going to wait to give @tixxit, @denisrosset, and @larsrh a chance to weigh-in, but I'm inclined to merge this.

denisrosset:
I'd like to add a test for the correctness fix, the following line would do:
```scala

 assert(p === Polynomial("1/4x² + x + x + 1/2"))
```

PR #413, opened 2015-03-23
Upgrade minor versions of sbt, scala and libraryDependencies

seratch:
This pull request upgrades minor versions of sbt, scala and libraryDependencies. 
I've checked dependencies' latest versions by using [sbt-updates](https://github.com/rtimush/sbt-updates).

- Bump sbt 0.13.7 to 0.13.8
- Bump Scala minor versions (also Travis settings)
- Bump scalatest, scalacheck, caliper stuff and so on
- Bump sbt-pgp 0.8 to 1.0.0 (I'm already heavily using in my projects)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: scala
 scala:
-  - 2.10.2
-  - 2.11.2
+  - 2.10.5
+  - 2.11.6
 notifications:
   irc:
     channels:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: scala
 scala:
-  - 2.10.5
+  # https://github.com/non/spire/pull/413#issuecomment-89896773
+  - 2.10.2
   - 2.11.6
 notifications:
   irc:
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -54,7 +54,8 @@ object MyBuild extends Build {
 
     scalaVersion := "2.11.6",
 
-    crossScalaVersions := Seq("2.10.5", "2.11.6"),
+    // https://github.com/non/spire/pull/413#issuecomment-89896773
+    crossScalaVersions := Seq("2.10.2", "2.11.6"),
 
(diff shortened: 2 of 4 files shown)


non:
Seems like a transient build error. I'm re-running, will merge if it passes.

Thanks!

non:
It seems like the build keeps stalling and getting killed by Travis.

Have you completed a clean build? Any sense of how long you expect it to take? It's possible one of these upgrades extended the duration in 2.10 until the build doesn't finish on time.

seratch:
I just tried `sbt ++2.10.5 clean test` again and that successfully finished in 493 seconds on my machine. Would you rerun the Travis job?

non:
So at this point I've run the build 4-5 times and it's failed every time.

It seems like we're just exceeding Travis' maximum time allowed without output.

non:
I'll try to look into possible ways to speed up the build under 2.10. I'm nervous about giving up on 2.10 in Travis, since it has caught real compatibility bugs. I guess we'll see.

seratch:
Thanks. After doing `rm -rf */target`, this problem represented on my machine too. I'll investigate more on that.

non:
Hey there,

Sorry to possibly conflict with you on #419 -- this was something I was working on before this PR came in. If that gets merged I can try to help update this PR to get it merged as well.

seratch:
just rebased against the master branch although I've not fixed the Scala 2.10 build issue yet..

non:
In the short term, you might want to try reverting to 2.10.2 and see if the rest of it works.

We chose not to upgrade to 2.10.3 or 2.10.4 because the compile times were way too slow, and there weren't any obvious problems we noticed with 2.10.2. So I'd be fine with leaving it back here as well.

seratch:
Thanks. I just reverted Scala 2.10 minor version and Travis build looks fine. https://github.com/seratch/spire/commit/7df155e1c1a254f7f5d4c3984347005db8fdb35a

non:
Merging. Thanks!

PR #418, opened 2015-04-05
spire.random.Op#resume should be final.

xuwei-k:
scalac can't apply tail call optimization non-final method.


non:
Wow, good catch. Thanks!

PR #419, opened 2015-04-06
Use stricter scalac options with Spire.

non:
This commit adds several extra scalac options that will help
us catch errors. All warnings have been eliminated, and moving
forward we should prevent any new warnings from being added.

(Due to inliner warnings, which can't be suppressed, we can't
use -Xfatal-warnings, unfortunately.)

Eventually I'd like to do more static checking (and maybe
enforce some common style/idioms), but this is a good first
step.

### DIFF
--- a/benchmark/src/main/scala/spire/benchmark/AnyValAddBenchmarks.scala
+++ b/benchmark/src/main/scala/spire/benchmark/AnyValAddBenchmarks.scala
@@ -25,7 +25,7 @@ class AnyValAddBenchmarks extends MyBenchmark {
   var floats:Array[Float] = null
   var doubles:Array[Double] = null
 
-  override protected def setUp() {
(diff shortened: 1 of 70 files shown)


InTheNow:
@non As part of https://github.com/scala-js/scala-js/pull/1549, I have had to explicitly add `: Unit` to method definitions as you have done here. 

Furthermore, there was also the issue of calling such methods "with (), since it is a side-effecting function" - for example [CForBenchmark](https://github.com/non/spire/blob/af1ccb874fdc2fbb8b2c52da64bad50cbb7a7777/benchmark/src/main/scala/spire/benchmark/CForBenchmark.scala#L33) -> `def timeWhileOr(reps:Int) = run(reps)(doWhileOr())`. 

I would _dearly_ love to help out here but I'm still bogged down with sjs :)

PR #422, opened 2015-04-10
Remove ApproximationContext[Rational] and all uses.

### LINKED ISSUE #392: Bug in Rational nroot
Rational.nroot or pow hangs for some very simple test cases. E.g. 

```scala
implicit val ac = ApproximationContext(Rational(1L, 100000000000000000L) // I can also set this to 1, it does not seem to matter
Rational(3) pow Rational(2) // hangs
```

The cause seems to be that Rational.nroot(Long,Int) does not terminate for some simple inputs. It probably just misses an additional cutoff.

```scala
Rational.nroot(9,1) // hangs
```

non:
This commit tears out support for NRoot[Rational] using
implicit ApproximationContext[Rational] instances. Now
that we have Algebraic and Real, we don't need to support
this case.

We do still have Numeric[Rational] and Fractional[Rational],
but those just use Double approximations. Users who require
more precision can reach for Real or Algebraic.

Fixes #392.

### DIFF
--- a/core/src/main/scala/spire/math/ApproximationContext.scala
+++ b/core/src/main/scala/spire/math/ApproximationContext.scala
@@ -1,6 +1,6 @@
 package spire.math
 
 case class ApproximationContext[A](error: A)
-object ApproximationContext {
-  implicit def rational2error(q: Rational) = ApproximationContext(q)
-}
+// object ApproximationContext {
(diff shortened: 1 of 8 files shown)


non:
Let's leave this open -- I'm going to try to remove all uses of `ApproximationContext`.

tixxit:
:+1:

PR #425, opened 2015-04-18
fix scalacOptions, remove unused imports

xuwei-k:
`scalacOptions in (Compile, compile)` does not work
http://stackoverflow.com/a/27858944


non:
Looks great! Going to merge this manually. Thanks! :+1:

Issue #424, opened 2015-04-17
error: not found: value Cmwc5


kevinmeredith:
I got a compile-time error when following the first example from: https://github.com/non/spire/blob/master/README.md#pseudo-random-number-generators

```
scala> import spire.math._
import spire.math._

scala> import spire.random._
import spire.random._

scala> val rng = Cmwc5()
<console>:25: error: not found: value Cmwc5
       val rng = Cmwc5()
                 ^
```

kevinmeredith:
Adding `import spire.random.rng._` enabled creation of `rng`, but failed on implicit look-up:

```
scala> import spire.math._
import spire.math._

scala> import spire.random.rng._
import spire.random.rng._

scala> val rng = Cmwc5()
rng: spire.random.rng.Cmwc5 = spire.random.rng.Cmwc5@2e17d83d

scala> rng.next[Double]
res0: Double = 0.2760649393497461

scala> rng.next[Complex[Double]]
<console>:15: error: could not find implicit value for parameter next: spire.random.Dist[spire.math.Complex[Double]]
              rng.next[Complex[Double]]
                      ^
```.

I'm looking into it - assuming that's OK.

**Comment by kevinmeredith** [comment]
So `Cmwc5.scala` extends `LongBasedGenerator`, which extends `Generator`.

`Generator#next` has the body:

`def next[A](implicit next: Dist[A]): A = next(this)`

I attempted to resolve the implicit by importing [spire.random.Dist.complex](https://github.com/non/spire/blob/master/core/src/main/scala/spire/random/Dist.scala#L295), but that failed:

```
scala> import spire.random.Dist.complex
import spire.random.Dist.complex

scala> rng.next[Complex[Double]]
<console>:20: error: could not find implicit value for parameter next: spire.random.Dist[spire.math.Complex[Double]]
              rng.next[Complex[Double]]
                      ^
```

**Comment by non** [comment]
I think the real issue here is that the `Dist[Complex[A]]` instance requires some type classes for `A`, such as `Trig[A]` and `Fractional[A]`.

Try this:

```scala
import spire.implicits._
import spire.math._
import spire.random._
val r = rng.Cmwc5()
r.next[Complex[Double]]
```

kevinmeredith:
Closing per https://github.com/non/spire/pull/426

PR #426, opened 2015-04-19
Add Import for Implicit Resolution in README

kevinmeredith:
Addresses #424 per @non's [fix](https://github.com/non/spire/issues/424#issuecomment-94286455).


non:
Thanks!

PR #427, opened 2015-04-20
WIP: Algebraic Refactor (+ rational root isolation/refinement)

tixxit:
This is a WIP PR. I still need to write some more tests. I was hoping to get more written, but keep getting bogged down in actually fixing bugs the tests find :)

This is a pretty big refactoring of `Algebraic`. It simplifies the whole structure greatly (no more cake) and also makes optimizations a bit easier (less indirection). It also adds a new rational root constant type to Algebraic (eg. `Algebraic.roots(poly)` or `Algebraic.root(poly, 0)`).

This also adds algorithms and a bit of an API for rational polynomial root isolation and refinement (just `BigDecimal` refinement right now). Both use good algorithms; VAS for root isolation and QIR for refinement.

For the API, if you have a `poly: Polynomial[BigDecimal]` then you can `val roots: Roots[BigDecimal] = poly.roots(MathContext.DECIMAL128)`, for instance. From there, `roots(0)`, `roots.count`, etc. The `roots` method actually takes an implicit `RootFinder`, which is a bit weird. So, there are implicit conversions from `MathContext => RootFinder[BigDecimal]`, for instance, which is why the previous example works. It also means if we have a type like `Double`, we can just call `poly.roots` since `RootFinder[Double]` wouldn't require any context (eg MathContext), like BigDecimal does.


non:
Thanks so much!

If there is any simple testing you'd like help with, feel free to comment here and I can submit PRs against this branch. Also, at any point where you feel like this is a net improvement over what we have, I'd be happy to merge it (which will probably make collaborating on bug fixes easier).

non:
By the way, some warnings (as errors) under 2.10:

```
[error] /home/travis/build/non/spire/core/src/main/scala/spire/math/Algebraic.scala:288: reference to mc is ambiguous;
[error] it is both defined in method toBigDecimal and imported subsequently by 
[error] import Expr._
[error]     val roundingMode = mc.getRoundingMode
[error]                        ^
[error] /home/travis/build/non/spire/core/src/main/scala/spire/math/Algebraic.scala:336: reference to mc is ambiguous;
[error] it is both defined in method toBigDecimal and imported subsequently by 
[error] import Expr._
[error]     val approx = rec(expr, mc.getPrecision + 2)
[error]                            ^
[error] /home/travis/build/non/spire/core/src/main/scala/spire/math/Algebraic.scala:337: reference to mc is ambiguous;
[error] it is both defined in method toBigDecimal and imported subsequently by 
[error] import Expr._
[error]     val newScale = approx.scale - approx.precision + mc.getPrecision
[error]                                                      ^
[error] /home/travis/build/non/spire/core/src/main/scala/spire/math/Algebraic.scala:342: reference to mc is ambiguous;
[error] it is both defined in method toBigDecimal and imported subsequently by 
[error] import Expr._
[error]       .round(mc) // We perform a final round, since roundExact uses scales.
[error]              ^
[error] four errors found
[error] (core/compile:compileIncremental) Compilation failed
[error] Total time: 110 s, completed Apr 20, 2015 3:58:29 PM
```

tixxit:
Ah shoot - teach me for not `+test`ing! FWIW, I think it is pretty much a net improvement over what existed already. Likely many areas that should be ~easy to test if you're feeling up for it, so I'll go through the PR a bit and comment on a few places that are probably bug-prone.

core/src/main/scala/spire/math/Algebraic.scala
```diff
@@ -66,86 +167,1383 @@ extends ScalaNumber with ScalaNumericConversions
...
-  override def hashCode: Int = if (isWhole && toBigInt == toLong) {
+  override def hashCode: Int = if (isWhole && isValidLong) {
     unifiedPrimitiveHashcode
   } else {
     val x = toBigDecimal(java.math.MathContext.DECIMAL64)
     x.underlying.unscaledValue.hashCode + 23 * x.scale.hashCode + 17
   }
-}
 
+  def toExprString: String = {
```
tixxit: I'll delete this. It's nice to have something like this, but this particular implementation is janky and I'd rather not support it (or test it in this PR).

core/src/main/scala/spire/math/Algebraic.scala
```diff
@@ -66,86 +167,1383 @@ extends ScalaNumber with ScalaNumericConversions
...
+   * index a real root (eg the value is greater than or equal to the number of
+   * real roots) then an `ArithmeticException` is thrown. Roots are indexed
+   * starting at 0.  So if there are 3 roots, then they are indexed as 0, 1,
+   * and 2.
+   *
+   * @param poly the polynomial containing at least i real roots
+   * @param i    the index (0-based) of the root
+   * @return an algebraic whose value is the i-th root of the polynomial
+   */
+  def root(poly: Polynomial[Rational], i: Int): Algebraic = {
```
tixxit: Should add some tests to verify that calling this with an invalid index (< 0 or >= # of roots) throws an exception.

core/src/main/scala/spire/math/Algebraic.scala
```diff
@@ -1,64 +1,165 @@
...
+    this - (this /~ that) * that
+
+  /** Returns the square root of this number. */
+  def sqrt: Algebraic = nroot(2)
+
+  /** Returns the cube root of this number. */
+  def cbrt: Algebraic = nroot(3)
+
+  /** Returns the `k`-th root of this number. */
+  def nroot(k: Int): Algebraic = if (k < 0) {
```
tixxit: We should probably test this with very large values (namely > 306) and make sure it's OK. I'm not sure how slow this will be. I think just testing this with constants (or very simple expressions) should be fine for now, since we're not testing how the underlying errors propagate.

core/src/main/scala/spire/math/Algebraic.scala
```diff
@@ -1,64 +1,165 @@
...
+    new Algebraic(Expr.Div(Expr.ConstantLong(1), Expr.KRoot(this.expr, -k)))
+  } else if (k > 0) {
+    new Algebraic(Expr.KRoot(this.expr, k))
+  } else {
+    throw new ArithmeticException("divide by zero (0-root)")
+  }
+
+  /** Raise this number to the `k`-th power. */
+  def pow(k: Int): Algebraic =
```
tixxit: Would be nice to test this with large values (performance permitting) and also some of the exceptional cases (0, Int.MinValue, etc).

core/src/main/scala/spire/math/Algebraic.scala
```diff
@@ -66,86 +167,1383 @@ extends ScalaNumber with ScalaNumericConversions
...
+  /**
+   * Relative approximation to the precision specified in `mc` with the given
+   * rounding mode. Rounding is always exact. The sign is always correct; the
+   * sign of the returned `BigDecimal` matches the sign of the exact value this
+   * `Algebraic` represents.
+   *
+   * @param mc the precision and rounding mode of the final result
+   * @return an approximation to the value of this algebraic number
+   */
+  def toBigDecimal(mc: MathContext): BigDecimal = {
```
tixxit: It would be nice to have property tests that verify the claim that the rounding is precise for all algebraic numbers. Not sure the best way to go about this though. Could also try just adding a known numbers + a very small number and verify that rounding is correct. Not sure...

non: This might be a case where we can use `Real` and `Algebraic` to test each other?

non:
:+1: 

Once Travis passes I'm going to merge this -- it's a big improvement over the existing thing. Any issues with this PR can just become Spire issues.

PR #438, opened 2015-04-28
Adding Generic PositiveInt Macro & Apply to JetDim's > 0 `dimension` Field

kevinmeredith:
https://github.com/non/spire/issues/437


kevinmeredith:
Posted this SO question due to difficulty using the macro - http://stackoverflow.com/questions/29949563/using-macro-to-make-case-class

kevinmeredith:
This compiled and tested locally for me. Thanks to @tixxit, who told me exactly how to implement the macro in my above SO question.

Aside: Here's follow-up discussion: https://groups.google.com/forum/#!topic/spire-math/FYWFkcZI7Ic.

I think this is ready for review. But I need to look at how to test macros - I hadn't used them before this PR, and I don't know how to test them.

kevinmeredith:
I double-checked that there's whitespace at the end of each file.

I suppose that adding tests, i.e. to ensure that the `PositiveInt` macro and its usage in `JetDim` work correctly, is preventing this PR's acceptance?

kevinmeredith:
After merging `upstream/master`:

```
scala> import spire.math._
import spire.math._

scala> JetDim(100)
res1: spire.math.JetDim = JetDim(PositiveInt(100))

scala> JetDim("-55".toInt)
java.lang.IllegalArgumentException: requirement failed
```

With respect to unit testing macros, is there any update to [SI-5989](https://issues.scala-lang.org/browse/SI-5989)?

kevinmeredith:
I saw the following failure in the Travis CI [Build](https://travis-ci.org/non/spire/jobs/76397628):

>org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed

I'm not sure if my change had caused it, but I'm going to close & re-open this PR to trigger a re-build.

rklaehn:
To rerun travis CI, you don't have to close and reopen the PR. There is a button on the CI gui (on the right of the header panel) to rerun the CI. Sadly, this is necessary sometimes.

Since the CI runs very long, I usually run
```
sbt validateJVM
sbt ++2.10.5 validateJVM
```
locally.

rklaehn:
I reran the CI, and this time both 2.10 and 2.11 passed.

kevinmeredith:
Thanks @rklaehn. Once I figure out how to test macros (http://stackoverflow.com/questions/32173578/testing-a-macro), I'll add them for this macro.

kevinmeredith:
Ran into this [issue](https://groups.google.com/forum/#!topic/scalatest-users/Mswf_wRk4xk) when trying to test a simple macro.

Once this [change](https://github.com/scalatest/scalatest/pull/669) gets published, I'll update the `build.sbt` to reference that new scalatest version. And, I'll add the macro test.

kevinmeredith:
I think that there's a mistake in the above macro - demonstrated by this [question](https://groups.google.com/forum/#!topic/scala-language/AI3wydstm-g).

kevinmeredith:
Both of the following ran successfully for me:

```
sbt validateJVM
sbt ++2.10.5 validateJVM
```

I don't see how to manually re-build, so I'm closing and re-opening.

-http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

kevinmeredith:
Hmm. In both builds I'm seeing the following failure. 

```
[info] Fast optimizing /home/travis/build/non/spire/macros/.js/target/scala-2.11/spire-macros-test-
fastopt.js
java.io.IOException: Failed to deserialize info of 
/home/travis/.ivy2/cache/org.scalatest/scalatest_sjs0.6_2.11/bundles/scalatest_sjs0.6_2.11-3.0.0-
M8.jar#org/scalatest/prop/Tables.sjsir
```

This isn't related to my changes, is it? Could you please take a look, @rklaehn?

Thanks.

kevinmeredith:
Could you please take a look, @non - https://github.com/non/spire/pull/438#issuecomment-140604839?

Thanks!

tixxit:
Given that we aren't using `primitives` in this PR, could we get rid of them in this PR and make a separate PR to review them separately?

tixxit:
Any implications to making the `PositiveInt` class a value class instead of a case class?

tixxit:
... other than the public constructor, which sucks but I could live with.

non:
Thanks for the gentle poke @kevinmeredith :)

First of all, sorry I haven't been able to give this PR as much love as it deserved. I'm going to see if I can get Travis to successfully build it (it doesn't seem like the errors have anything to do with your code).

The macros all look fine, but I see a few issues. It looks like we will unconditionally be doing a `require(n > 0)` check at runtime (even when the macro has verified this at compile-time). Also, it seems like we will be allocating an instance of the class instead of using a value class. Ideally there should be little-or-no overhead when using this with an `Int` constant.

What do you think? Could we move the require check into a factory constructor that gets bypassed when called from a macro? Could we make this a value class? Also, worth adding `NonNegative` and `NonZero` wrappers too?

non:
@tixxit Yeah, if it's a value class then folks will always be able to bypass the safety by instantiating it directly, but I can live with that if we clearly document how we expect the type to be used.

tixxit:
What do you think of having an alternative implicit `Int => PositiveInt` conversion (macro) that *requires* a literal to compile. That way, people could still just `Semigroup.combineN(x, 23)`.

non:
@tixxit that works for me.

So to be clear, the design would be:

 * implicit `Int => PositiveInt` using macro, requiring literal value, no runtime check
 * explicit `PositiveInt(n)` factory constructor, with runtime check
 * `new PositiveInt(n)` unsafe constructor without check whose use is discouraged.

Is this what you mean?

tixxit:
Yes. Though having the `PositiveInt(n)` constructor be a macro that can optionally elide the runtime check is nice too. I feel like some folks may not like the implicit conversion and just use `PositiveInt(n)` for cleanliness and I don't want to punish them for it.

non:
Right, OK, so let's amend the factory constructor to "use macro if able, otherwise runtime check"

non:
So I think we can fix the build error by updating to scala.js 0.6.5 (at least, that worked locally after things failed for me).

non:
@kevinmeredith I'm going to push your branch to `origin` with a proposed extra commit. I'll link it here and see what you think.

non:
Take a look at 65f0d4192 and see what you all think.

non:
@kevinmeredith Would you like me to open a PR based on the branch with that commit? If we did that, then @tixxit, @rklaehn, or someone else can sign-off on it and we can merge it into master.

kevinmeredith:
> Would you like me to open a PR based on the branch with that commit?

Yes, please, @non.

kevinmeredith:
Perhaps it's worthwhile for this PR to close, and then use `shapeless`'s `Nat`?

Note - I've never used this library's `Nat`, but I imagine it's the same as Idris's.

I believe that @tixxit had suggested this route previously.

Issue #431, opened 2015-04-28
Algebraic: review/improve use of Rational/BigInt in root isolation & refinement


tixxit:
The current methods to do root isolation and refinement make heavy use of BigInt and Rational. The problem is that for a value like 23e1000 we are wasting a lot of space to, basically, store some 0s. It would be nice to see if we can use a wrapper around BigDecimal here instead, which guarantees the value is an Int. Or perhaps some hybrid type of Rational/BigDecimal (like Number?).

PR #442, opened 2015-05-03
Use BigDecimal arithmetic in BigDecimalRootRefinement.

tixxit:
This gets rid of `BigInt` and `Rational` arithmetic used in `BigDecimalRootRefinement`. The main worry being that very large/small exponents would force us to use a lot of memory unncessarily by forcing them to `BigInt`s or `Rational`s.

This is a partial fix for #431 - which handles the refinement side (but not isolation).


non:
Looks good to me. :+1:

PR #443, opened 2015-05-04
Add === for Types

kevinmeredith:
https://github.com/non/spire/issues/423

### DIFF
--- a/core/src/main/scala/spire/math/FixedPoint.scala
+++ b/core/src/main/scala/spire/math/FixedPoint.scala
@@ -42,6 +42,9 @@ class FixedPoint(val long: Long) extends AnyVal { lhs =>
     else throw new FixedPointOverflow(long)
 
   def === (rhs: FixedPoint): Boolean = lhs.long == rhs.long
+
+  def =!= (rhs: FixedPoint): Boolean = !(this === rhs)
+
   def != (rhs: FixedPoint): Boolean = lhs.long != rhs.long
 
   def abs: FixedPoint =
--- a/core/src/main/scala/spire/math/FixedPoint.scala
+++ b/core/src/main/scala/spire/math/FixedPoint.scala
@@ -41,7 +41,7 @@ class FixedPoint(val long: Long) extends AnyVal { lhs =>
     if (long != Long.MinValue) new FixedPoint(-long)
     else throw new FixedPointOverflow(long)
 
-  def == (rhs: FixedPoint): Boolean = lhs.long == rhs.long
+  def === (rhs: FixedPoint): Boolean = lhs.long == rhs.long
   def != (rhs: FixedPoint): Boolean = lhs.long != rhs.long
 
   def abs: FixedPoint =
--- a/core/src/main/scala/spire/math/Real.scala
+++ b/core/src/main/scala/spire/math/Real.scala
@@ -28,15 +28,15 @@ sealed trait Real extends ScalaNumber with ScalaNumericConversions { x =>
   override def hashCode(): Int = toRational.hashCode
 
   override def equals(y: Any): Boolean = y match {
-    case y: Real => ===(y)
+    case y: Real => this === y
     case y => toRational.equals(y)
   }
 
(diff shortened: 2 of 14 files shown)


kevinmeredith:
I believe that this PR is ready for review, please.

core/src/main/scala/spire/math/Real.scala
```diff
@@ -28,10 +28,14 @@ sealed trait Real extends ScalaNumber with ScalaNumericConversions { x =>
   override def hashCode(): Int = toRational.hashCode
 
   override def equals(y: Any): Boolean = y match {
-    case y: Real => x eqv y
+    case y: Real => ===(y)
     case y => toRational.equals(y)
   }
 
+  // TODO: should `eqv` and `===` both be used?
```
non: Good catch! Let's replace `eqv` with `===`.

non:
One larger comment: when calling the method, let's write `this === that` instead of `===(that)`. I think the former is a lot easier for people to visually-parse.

non:
Another thought I just had: do you think we should add a `=!=` method? otherwise, someone will need to write `!(x === y)`. What do you think?

kevinmeredith:
>Another thought I just had: do you think we should add a =!= method? 

Definitely. Please let me know any improvements/comments/suggestions/criticisms.

I have to admit that one of my interests in contributing to non/spire is to learn from you especially (and maybe contribute/learn more about non/cats).

core/src/main/scala/spire/math/ULong.scala
```diff
@@ -64,6 +64,7 @@ class ULong(val signed: Long) extends AnyVal {
     if (signed >= 0L) signed.toString
     else (ULong.LimitAsBigInt + signed).toString
 
+  // TODO: replace these with `===` and `=!=`? Don't we get `==` and `!=` for free with value classes?
```
kevinmeredith:
```
scala> class LongWrapper(val x: Long) extends AnyVal {}
defined class LongWrapper

scala> new LongWrapper(50)
res7: LongWrapper = LongWrapper@32

scala> new LongWrapper(50)
res8: LongWrapper = LongWrapper@32

scala> res7 == res8
res9: Boolean = true

scala> res7 != res8
res10: Boolean = false
```

non: Ah, so maybe we need to keep `==` and `!=` as well.

kevinmeredith: My incomplete understanding is that we get `==` and `!=` for free with a value class (like the above example). That's just an observation from my **minimal** usage of value classes.

If that's true, then, regardless of these `===` and `=!=` methods, why would `==` and `=!` need to be defined?

**Caveat** - I don't actually how know value classes' `==` and `!=` get implemented, such as in the above `LongWrapper `example.

kevinmeredith:
Hmm. `NumberTest` is failing:

```
[info] - operations *** FAILED ***
[info]   9223372036854775806 equaled 9.223372036854776E18 (NumberTest.scala:131)
```

Looking into it.

core/src/main/scala/spire/math/FixedPoint.scala
```diff
@@ -41,7 +41,10 @@ class FixedPoint(val long: Long) extends AnyVal { lhs =>
     if (long != Long.MinValue) new FixedPoint(-long)
     else throw new FixedPointOverflow(long)
 
-  def == (rhs: FixedPoint): Boolean = lhs.long == rhs.long
+  def === (rhs: FixedPoint): Boolean = lhs.long == rhs.long
+
+  def =!= (rhs: FixedPoint): Boolean = !(this === rhs)
+
   def != (rhs: FixedPoint): Boolean = lhs.long != rhs.long
```
kevinmeredith: Should this `!=` be removed too?

non:
Hey, sorry I dropped the ball on this. I think we should merge it as-is now and then fix up any problems in a future PR. I'm going to try to fix the merge conflicts locally and merge this now.

PR #445, opened 2015-05-06
WIP - Add continued fraction type.

kevinmeredith:
addresses https://github.com/non/spire/issues/234

### DIFF
--- a/core/src/main/scala/spire/math/ContinuedFraction.scala
+++ b/core/src/main/scala/spire/math/ContinuedFraction.scala
@@ -1,0 +1,37 @@
+package spire.math
+
+import scala.annotation.tailrec
+
+object ContinuedFraction {
+	
+	def apply(r: Rational): ContinuedFraction = 
+		ContinuedFraction( computeFirst(r) )
+
+	// would prefer to use scalaz.NonEmptyList here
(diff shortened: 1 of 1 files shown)


non:
Hi @kevinmeredith!

Sorry! This is my fault. I actually did a bunch of work on this, but hit a point where I wasn't sure it was worth it to actually include. I'm sorry I didn't do a good job of keeping the bug report up-to-date.

The branch I have is here:

https://github.com/non/spire/tree/topic/continued-fractions

Take a look at the `CF` type:

https://github.com/non/spire/blob/topic/continued-fractions/core/src/main/scala/spire/math/CF.scala

non:
(Basically, I lost confidence that `CF` had any major capabilities that `Real` lacked, and reached the edge of my ability to work with continued fractions. For example, I wanted to implement rational roots of continued fractions, but ended up putting it down before figuring out how best to do that.

kevinmeredith:
Cool! Thanks for pointing me to that branch. 

I look forward to reading your code for CF!

non:
Here's Bill Gosper describing the approach that this code uses: http://perl.plover.com/yak/cftalk/INFO/gosper.txt

I was turned on to the idea by Mark Dominus: http://perl.plover.com/yak/cftalk/

PR #446, opened 2015-05-07
removed superflous "min" parameter from Interval:

siddhartha-gadgil:
at core/src/main/scala/spire/math/Interval.scala


denisrosset:
Seems that the CI triggered a bug in the rational root finder. Restarting the CI test.

```
[info] - relative approximation of zero is zero
[info] - find all rational roots of rational polynomial *** FAILED ***
[info]   RuntimeException was thrown during property evaluation.
[info]     Message: invalid isolated root interval: ((Ø),0)
[info]     Occurred when passed generated values (
[info]       arg0 = List(-85/3536609, -2161016731/2130797184) // 1 shrink
[info]     )
```

non:
Yeah I just restarted it again.

If this one fails too let's just merge this to master (since the issue is already in master) and look at fixing the rational root finder there.

PR #447, opened 2015-05-07
Use Typelevel mailing list for spire

pleira:
As recommended in this [post](https://groups.google.com/forum/#!topic/spire-math/cNMuOxSEKHA).


non:
Thanks!

PR #454, opened 2015-05-27
Enables and configures scalastyle

rklaehn:
Adds a scalastyle-config.xml with relatively lenient settings
Removes all trailing whitespaces
Adds a lot of explicit return types
Disables some scalastyle warnings explicitly (very long methods etc.)

There are some equals/hashcode warnings that are intentionally not disabled since I think there might be something to them.


non:
Thanks!

:+1:  once Travis passes.

PR #457, opened 2015-05-28
Adds law checks for some non-abstract typeclass methods

rklaehn:
I am sure there are more. But these are all I could find with a quick look.


non:
:+1: once Travis passes.

non:
Thanks!

PR #458, opened 2015-05-30
Added Merging trait and two implementations

rklaehn:
This is an algorithm to merge two arrays with a minimum number of comparisons. It is usually faster than the naive linear merge implementation even for types with a cheap compare operation such as Int, because it copies large segments using System.arraycopy. For types with a comparatively expensive comparison operation, such as Rational or String, the gain is even larger.

This also adds tests and checks. Test coverage for the new code should be 100%
There is also a small Thyme-based benchmark that shows the advantage over a linear merge. I am not sure what is the best place to put thyme-based benchmarks.

### DIFF
--- a/benchmark/src/main/scala/spire/benchmark/MergeBenchmark.scala
+++ b/benchmark/src/main/scala/spire/benchmark/MergeBenchmark.scala
@@ -1,0 +1,22 @@
+package spire.benchmark
+
+import spire.implicits._
+import spire.math.Rational
+import spire.math.BinaryMerge
+import spire.math.LinearMerge
+
+import ichi.bench.Thyme
+
+object MergeBenchmark extends App {
+
+  val th = Thyme.warmed(verbose = println)
+  // val th = new Thyme()
+
+  val ar = (0 until 1000).map(Rational.apply).toArray
+  val br = (1000 until 1001).map(Rational.apply).toArray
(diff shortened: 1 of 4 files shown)


rklaehn:
We might want to move this to spire-extras once that exists.

non:
Looks good to me! :+1: 

Honestly, this is the sort of thing I think folks might widely benefit from.

So I'm happy to keep it in `core` for now. Eventually we might want to move sorting/searching/selecting/merging into their own project. I see spire-extras as being more about exotic types, rather than these kinds of utilities/algorithms.

(I will leave the PR open to give @tixxit, @denisrosset, and @larsrh a chance to comment but unless I hear complaints soon I will merge it.)

rklaehn:
I am going to use this from the IntervalSeq implementation (rklaehn/intervalseq currently has its own version of this). I also think you could do a very nice unboxed immutable Set or Seq based on this. I would not even bother trying to link it into the scala collections hierarchy though.

non:
Hey I'm really sorry I failed to merge this awhile ago.

I like the design, so I'm going to merge this. But in the future I'd like to rename `trait Merge` to be `trait Merging` to match the file name (and the convention with other types, `Sorting`, `Searching`, etc). Is this OK?

rklaehn:
Feel free to rename the trait.

Issue #214, opened 2014-03-11
Add more random distributions to spire.random


non:
We have support for `Uniform` and `Gaussian` currently.

It would be nice to get support for at least: binomial, poisson, gamma, exponential. This might be a rabbit hole but being able to specify which distribution you want, rather than just getting a `Random[A]` instance, seems worth it.

PR #460, opened 2015-06-05
New spire.stats package

andrelfpinto:
Hi,
I thought about an alternate proposal for the `random` package as seen in #214.
In the present structure, `random.Dist` is a way of producing arbitrary values from a `Generator`. However a probability distribution should be a way of assign a probability to each measurable subset of the possible outcomes of a random experiment [1], i.e. you should be able to ask it for pdf, cdf, mean etc. The mechanism for production of values according to a probability distribution function should in reality be a random variable [2].
With the proposed separation it should be possible to characterize a probability distribution and still be able to modify the production of values from the `Generator` (the role of `random.Dist` would basically go to `stats.RandomVariable`).
`stats.Distribution` and `stats.RandomVariable` would be immutable.
Maybe it's possible to have some random variable `Z` to be a combination of random variables `X` and `Y`, as `Z = X + Y`.
Anyway I made a few scratches but I would like your blessing ;) to see if this is a good strategy.

[1] http://en.wikipedia.org/wiki/Probability_distribution
[2] http://en.wikipedia.org/wiki/Random_variable

### DIFF
--- a/core/src/main/scala/spire/stats/Distribution.scala
+++ b/core/src/main/scala/spire/stats/Distribution.scala
@@ -1,0 +1,33 @@
+package spire.stats.distribution
+
+import spire.random.Generator
+
+trait Distribution[A] {
+
+  /** Function used to transform the values drawn from a Generator according to this probability distribution function
+    */
+  private[stats] val transform: (Generator => A)
+}
+
+trait ContinuousDistribution[A] extends Distribution[A] {
+
+  /** Returns the probability density function (pdf) of this distribution at the value x
+    */
+  def pdf(x: A): A
+
+  /** Returns the cumulative distribution function (cdf) of this distribution at the value x
+    */
+  def cdf(x: A): A
+}
+
+trait DiscreteDistribution[A] extends Distribution[A] {
+
+  /** Returns the probability mass function (pmf) of this distribution at the value x
+    */
+  def pmf(x: A): A
+
+  /** Returns the cumulative distribution function (cdf) of this distribution at the value x
+    */
+  def cdf(x: A): A
+}
+
--- a/core/src/main/scala/spire/stats/Distribution.scala
+++ b/core/src/main/scala/spire/stats/Distribution.scala
@@ -31,3 +31,16 @@ trait DiscreteDistribution[A] extends Distribution[A] {
   def cdf(x: A): A
 }
 
+// Should a method that returns the next random value based on a
+// specified type be included in the Generator class itself?
+object RichGenerator {
+  import scala.reflect.runtime.universe._
+
+  implicit class RichGenerator(gen: Generator) {
+    def next0[A : TypeTag](): A = typeOf[A] match {
+      case t if t =:= typeOf[Double] => gen.nextDouble.asInstanceOf[A]
+      case _ => ???
+    }
+  }
+}
+
--- a/core/src/main/scala/spire/stats/Exponential.scala
+++ b/core/src/main/scala/spire/stats/Exponential.scala
@@ -11,9 +11,33 @@ trait ExponentialDistribution[A] extends ContinuousDistribution[A] {
 }
 
 object Exponential {
-  def Standard[A: Field: Trig: IsReal: TypeTag](): ExponentialDistribution[A] = new StandardExponentialDistributionImpl[A]()
+  def Standard[A: Field: Trig: IsReal: TypeTag: StandardBuilder](): ExponentialDistribution[A] = implicitly[StandardBuilder[A]].apply()
 
-  def apply[A: Field: Trig: IsReal: TypeTag](mode: A, rate: A): ExponentialDistribution[A] = new ExponentialDistributionImpl[A](mode, rate)
+  def apply[A: Field: Trig: IsReal: TypeTag: Builder](mode: A, rate: A): ExponentialDistribution[A] = implicitly[Builder[A]].apply(mode, rate)
+
+  abstract class StandardBuilder[A: Field: IsReal: TypeTag] {
+    def apply(): ExponentialDistribution[A]
+  }
+
+  abstract class Builder[A: Field: IsReal: TypeTag] {
+    def apply(mean: A, stdDev: A): ExponentialDistribution[A]
+  }
+
+  object StandardBuilder {
+    import spire.std.double._
+
+    implicit val doubleBuilder = new StandardBuilder[Double] {
+      def apply() = new StandardExponentialDistributionImpl()
+    }
+  }
+
+  object Builder {
+    import spire.std.double._
+
+    implicit val doubleBuilder = new Builder[Double] {
+      def apply(mean: Double, stdDev: Double) = new ExponentialDistributionImpl(mean, stdDev)
+    }
+  }
 
   class StandardExponentialDistributionImpl[A: Field: Trig: IsReal: TypeTag] extends ExponentialDistribution[A] {
     import spire.math._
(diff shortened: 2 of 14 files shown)


non:
@andrelfpinto Sorry I missed this earlier!

So, I think that a `spire.stats` package make sense, and I agree that we should support pdf, cdf, mean, etc (although obviously some distributions don't have a defined mean). However, I think there are a few things I'd like to change about this design.

First, I don't think we need the `RichGenerator` class, given that `Generator#next[A]` is already a (pluggable) mechanism for generating a new value based on a specified type. Ideally we could use this mechanism somehow (or update it to be consistent with this strategy). Do you imagine that you would port all the code in `Dist[A]` over to `RandomVariable[A]`? Also, do you imagine using it to implement `next0`?

Second, I like the more explicit treatment of distributions as first-class objects to work with (compared to the existing `spire.random.Gaussian`, `spire.random.Exponential`, etc). I can imagine operations to combine distributions. Currently I worry that the requirements are too strong (for example I'd like to be able to support uniform distributions of integers, which don't have a field). Another advantage of the current strategy is that different types can implement their own distribution classes (i.e. `Gaussian[Real]` and `Gaussian[Double]` don't necessarily have to share an underlying implementation).

When one knows the underlying pdf I think your strategy absolutely makes sense. However, in some cases it's not clear what the distribution is, but it is clear how to simulate it (and possibly observe the distribution through sampling) which is the way that `Dist[A]` (and also `Random[A]`) approach the problem. Ideally we would be able to support both approaches.

Thanks for your hard work, and sorry I couldn't respond earlier! Let me know if you have any questions about my comments, and what you think.

andrelfpinto:
Hi @non, don't worry about the timing.

As I said, it is just a few scratches (maybe a bit overly rough) and I knew the design didn't go well with the rest of the library. However I just needed it that way so I could get some feedback about wether it was a good idea to pursue or not. Also, I don't have the expertise you do with respect to type classes and I got a little confused when I started to make this code, so eventually I rip them off.

Sorry about the `RichGenerator`. It is a bit of stretch and ugly but when I was coding this, I did't want to mess with the original design. There is a `Generator#next[A]` method, but it needs an implicit `Dist[A]`. For most cases I could use that and pass an uniform distribution, but how about the uniform distribution itself? On the other hand, most PRNG algorithms (the generators) produce sequences which are uniformly distributed, so I could use a method that doesn't have to have an implicit `Dist[A]` parameter. Actually I argue that the burden to generate a new random number based on a specific distribution would lie in the `RandomVariable` class. The `RandomVariable` can have an underlying distribution and knows how to ask the distribution to give some random number, in other words, how to realize itself, giving "birth" to random variates. In the end, the generator used in the distributions needs to be an uniform distribution one (unless the distribution itself is implemented with some other type of generator in mind) and therefore I (humbly) think that `Generator#next[A]` should not have to receive an implicit `Dist[A]`.

I didn't go through all `Dist[A]` code in detail to know for sure if it would be possible to port it all to `RandomVariable[A]`, but since that in the design I propose `RandomVariable` has the same role of the original design's `Dist` (quoting you in #214: "Dist[A] is a way of producing arbitrary A values using a Generator"), I would guess yes.

Well I guess the confusion about the uniform distribution of integers is because of me. Maybe I wasn't explicit enough but I implemented the continuous uniform distribution as I wrote in the commit and I have not yet implemented the discrete one (that I would call `DiscreteUniformDistribution`). Maybe I should have called the other one `ContinuousUniformDistribution` but I thought it is more used so I shorten its name. As a side note, I discovered that the `DiscreteDistribution` trait (that would be used to implement it) must have to be changed due to fact that only one type is not enough in this case).

About the advantage of the current strategy that different types can implement their own distribution classes I also think that as an advantage but my implementation stopped in my inability to get the type classes working fully in my head. Maybe you have some more tips I could use in this case.

In the cases when we don't know the distribution I think we can use the `Distribution[A]` trait instead. The class `RandomVariableWithDistribution[A]` needs an `Distribution[A]` parameter, not one of the subclasses `ContinuousDistribution` or `DiscreteDistribution`. `Distribution[A]` has only one function to implement that is used to, quoting, "transform the values drawn from a Generator according to this probability distribution function". I guess the text is not that good since we don't need a pdf, we can implement it anyway we like it (actually, strictly speaking, the discrete distributions have a pmf). The way I see, in this design, we can support both approaches.

Thanks for your time to go through all the code, reviewing it, even as I said it was just some scratches.

core/src/main/scala/spire/stats/Distribution.scala
```diff
@@ -0,0 +1,46 @@
+package spire.stats.distribution
+
+import spire.random.Generator
+
+trait Distribution[A] {
+
+  /** Function used to transform the values drawn from a Generator according to this probability distribution function
+    */
+  private[stats] val transform: (Generator => A)
```
tixxit: Is there any reason this couldn't be a `RandomVariable[A]`?

tixxit:
First, I really like this PR in general - thanks!

```
About the advantage of the current strategy that different types can implement their own distribution classes I also think that as an advantage but my implementation stopped in my inability to get the type classes working fully in my head. Maybe you have some more tips I could use in this case.
```

So, the way the current strategy works is by having distribution constructors (eg. `random.Uniform`, `random.Gaussian`). So, `Uniform` doesn't give you a way to produce values, but rather gives you a way to create uniform distributions. So, you'd say: `Uniform[Int].apply(0, 100): Dist[Int]`. When you ask for an implicit `Uniform` the actual implementation can be chosen up front. Unfortunately, you'd have to do something kind of similar, since there isn't really a way to generalize "distribution constructors". I'm not sure if this is a huge burden though... shifting to your new code, may be something like:

```scala
trait Uniform[A] extends Distribution[A] {
  def min: A
  def max: A
}

object Uniform {
  def apply[A](min: A, max: A)(implicit builder: Builder[A]): Uniform[A] = builder(min, max)

  trait Builder[A] {
    def apply(min: A, max: A): Uniform[A]
  }

  object Builder {
    implicit val intBuilder = ...
    implicit val doubleBuilder = ...
  }
}
```

The builders cannot be generic (there will be a unique builder type class per distribution), but the distributions still can be generic.

Again, thanks for this.

core/src/main/scala/spire/stats/Distribution.scala
andrelfpinto: I guess it's more like a definition difference.
As seen in [1], a random variable is the set of outcomes from a random experiment, and the mathematical function describing the outcomes of the random variable and their associated probabilities is the (probability) distribution.
It's a composition relationship: a random variable has a probability distribution.
A probability distribution has a probability density function (or a probability mass function) and a cumulative distribution function. It can have location, scale or shape parameters, mean, etc.
The random variable has a distribution that you use to define it.

An example from [1]:
We can create a random variable `Y` that models a $1 payoff for a successful bet on heads as follows:

```
Y(w) = 1, if w=heads
       0, if w=tails
```

These are just the outcomes. Now, if the coin is a fair coin, `Y` has a probability mass function `f_Y` given by:

```
f_Y(y) = 1/2, if y=1
         1/2, if y=0
```

Also you can think about operating on random variables as: `Z = X + Y` but not on the distributions themselves [2].

My guess is that you are thinking: if the `transform` function is used only in the `RandomVariable[A]` why it's not defined there? Well, the reason I made it this way is that the `RandomVariable` just asks for whatever distribution it has to do the transformation since the distribution, when implemented, has all the information needed to do this. If we put it in the `RandomVariable` it would have to understand about the implementation.

[1] http://en.wikipedia.org/wiki/Random_variable
[2] http://math.stackexchange.com/questions/263974/sums-of-two-probability-density-functions

andrelfpinto:
I'll take a look and try to modify the code as you suggested. Thanks.

andrelfpinto:
Well, in the beginning I had a design similar to what this one is like now, using type classes. However I couldn't make it work the way it supposed to. Thank you for the explanation and the code, @tixxit, because now I had a certain that the direction to go was that one. It's actually pretty simple after one sees it done.

denisrosset:
Regrouped in umbrella issue.

Issue #452, opened 2015-05-26
Move SBT configuration to build.sbt


non:
At this point it seems like the format has improved, and the evolving standard for Scala libraries seems to be a single top-level SBT project.

For an example of this, see [Cats' build.sbt](https://github.com/non/cats/blob/master/build.sbt).

rklaehn:
Build is a build.sbt, so we might as well close this. Thanks @andrelfpinto

PR #462, opened 2015-06-12
Move SBT configuration to build.sbt

andrelfpinto:
Move SBT configuration to build.sbt #452


non:
Thanks! This looks great. :+1:

Issue #463, opened 2015-06-17
Add typeclass instances for Unit


rklaehn:
Sometimes it would be useful to have an Eq or Order instance for Unit. It is trivial to implement.

My use case for Eq[Unit] is a typeclass based map that I want to use as a set by setting the value type to Unit. I currently can't think of use cases for Unit instances of more complex typeclasses, but many of those would also be trivial to implement.

rklaehn:
The PR has been merged, so we might as well close this.

Issue #315, opened 2014-11-03
Spire style guide


denisrosset:
Spire/Scalaz/... are pioneering the use of typeclasses, however it is not immediately clear where to put the different parts of the typeclass and its syntax, and how to name them.

I would propose to create a small guide to help people find their way around the source code. Any ideas on the guidelines ?

This would also be helpful for libraries using the Spire typeclasses.
- the package structure is very well described in the [guide](https://github.com/non/spire/blob/master/GUIDE.md)
- when `math.X` is an object, there is a trait `XInstances` with the corresponding typeclass instances, and the companion object `math.X` extends `XInstances`,
- each type class instance is written in a separate trait. For "algebraic" classes (to be defined), a master trait `XAlgebra` collects all the pieces,
- these traits are instantiated in the `XInstances` trait using `implicit final val`, with the val name being the type name, thus emulating a singleton object.

(this variant is used by `UByte` and friends)
- or there is an implicit object in the `XInstances` trait collecting all the algebraic typeclasses.

(this variant is used by `SafeLong`)

This should be written along an explanation of the reasoning behind the structure.

Note: as the core maintainers form a small team, I don't think the project needs a strict guide, but it would be helpful for newcomers, as the typeclass pattern is not yet fully documented.

armanbilge:
I think this issue is pretty stale. Typeclasses are well established now and we can follow best practices from Cats, Cats Effect etc. Furthermore Scala 3 brings many improvements to this style of programming, so a Scala 2 focused style guide will have limited lifetime.

PR #464, opened 2015-06-19
Typeclass instances for Unit

rklaehn:
This provides some trivial but useful typeclass instances for Unit. So far I only implemented Eq[Unit] and Order[Unit], since these are the ones that I need.

I am not sure if I did the typeclass instance machinery correctly. #315 is not implemented yet, so I just looked at some of the other classes in spire/std and tried to copy things.

Implements #463 

Review by @non @tixxit

### DIFF
--- a/core/src/main/scala/spire/std/any.scala
+++ b/core/src/main/scala/spire/std/any.scala
@@ -17,3 +17,4 @@ trait AnyInstances extends BooleanInstances
     with MapInstances
     with ProductInstances
     with OptionInstances
+    with UnitInstances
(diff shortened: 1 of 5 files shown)


non:
Looks good. Can you add `object unit extends UnitInstances` to the `package.scala` object inside of std? That way, someone who wants to import their instances individually can get these.

non:
Looks good to me. :+1: 

I'm re-running Travis. When it passes I will merge.

tixxit:
Awesome :+1:

PR #465, opened 2015-06-23
Cleanup Roots a bit

tixxit:
Adds a default `toString` and extends `Iterable`.


non:
Could you also override the `size` method which `Iterable` gives to avoid traversing the entire iterator?

non:
:+1:

PR #469, opened 2015-07-12
Consistent unsigned unary negate

r0estir0bbe:
Currently `unary_-` of the Unsigned classes is not consistently defined, i.e. not all four classes have the same behaviour:
- [`UByte`](https://github.com/non/spire/blob/master/core/src/main/scala/spire/math/UByte.scala#L49) and [`UShort`](https://github.com/non/spire/blob/master/core/src/main/scala/spire/math/UShort.scala#L40) implement it by interpeting it as two's complement:

    ```scala
    def unary_- : U = U(-this.signed)
    ```
- [`UInt`](https://github.com/non/spire/blob/master/core/src/main/scala/spire/math/UInt.scala#L39) and [`ULong`](https://github.com/non/spire/blob/master/core/src/main/scala/spire/math/ULong.scala#L83) implement it by returning the same value:

    ```scala
    def unary_- : U = U(this.signed)
    ```

I fixed this by changing `UInt` and `ULong`'s implementation to match those of `UByte` and `UShort`. Additionally, I added test cases that check if `n + (-n) == 0` for all Unsigned classes.

### DIFF
--- a/core/src/main/scala/spire/math/UInt.scala
+++ b/core/src/main/scala/spire/math/UInt.scala
@@ -36,7 +36,7 @@ class UInt(val signed: Int) extends AnyVal {
   def >= (that: UInt): Boolean = this.toLong >= that.toLong
   def > (that: UInt): Boolean = this.toLong > that.toLong
 
-  def unary_- : UInt = UInt(this.signed)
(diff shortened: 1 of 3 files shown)


non:
:+1: Great work, thanks for catching this!

PR #471, opened 2015-07-13
Replace x -> y with (x, y).

non:
The "could not inline @inline-marked method ->$extension" message
is getting a bit old, and there's really no reason we need to be
using it.

For now, I left a few uses in the tests, but I would be happy to
remove those too.


tixxit:
:+1:

PR #472, opened 2015-07-18
This commit introduces sbt-sonatype.

non:
It also upgrades us to the latest sbt-release, which removes
a bunch of clunkiness from our SBT set up. We should now be
able to do non-interactive releases through Sonatype, which
makes cutting a release a much easier process.


non:
Oh I also forgot to mention the upgrade to Scala 2.11.7 from 2.11.6.

If you folks would prefer I can roll that back and do it in a separate PR.

PR #475, opened 2015-07-23
Rename scalacheck-binding to laws, etc.

non:
This commit simply moves the files from one place to another.
The laws are now located in a laws module/directory, and will
be published as spire-laws.

It also removes the `Arbitrary[_]` instances from ArbitrarySupport
(located in the tests project) and instead moves the relevant logic
to `Gen[_]` instances inside of spire.laws.gen and `Arbitrary[_]`
instances inside of spire.laws.arb.

It updates some of the instances to be a bit more idiomatic
and removes some potential problems. It also beings the work of
creating a wide variety of flexible `Gen[_]` instances to remove the
need to create so many ad-hoc `Arbitrary[_]` instances in our tests.

Finally, it begins standardizing around the SpireTests and SpireProperties
traits. These traits disable ScalaTest's `===` to avoid conflicts, and
start to standardize on the use of ScalaTest's matchers (e.g. `shouldBe`)
over the use of `assert()`.

(The difference between SpireTests and SpireProperties largely comes down to
whether you use `test("xyz")` or `property("xyz")`, and whether you are using
ScalaCheck to do property-based testing.)


non:
Review by @tixxit, @rklaehn, and @InTheNow.

non:
I had intended to completely standardize our tests (in terms of traits, package location, style, etc) before submitting a PR, but I realized that the results would be too big to review, and would hit conflicts with most future PRs. Hopefully a somewhat piecemeal approach will work to standardize our tests.

non:
That's a good point. I'll update it to use either the current approach or one based on `Double`.

rklaehn:
What's the purpose of disabling ScalaTest's ===? So you can use spires === instead?

non:
@rklaehn Yes, I would prefer if we could do `import spire.implicits._` (or similar) without having to worry about creating ambiguous implicits with ScalaTest.

non:
(If everyone else disagrees I can remove this part though.)

rklaehn:
Other than the rational gen, LGTM. I'm fine with disabling ScalaTests ===

non:
@rklaehn Your suggestion to improve `Gen[Rational]` exposed a couple of minor bugs (that I fixed). So obviously it was an insightful and timely suggestion. Thanks!

PR #476, opened 2015-07-23
Topic/add ===

non:
This is a version of #443 that has merge conflicts solved.

I also fixed a problem I noticed with `Number#===` and a missing `ULong#===`.

@kevinmeredith are you OK with these changes? If so, I'll go ahead and merge your commits.


non:
(Also, apologies again to @kevinmeredith for letting this sit. I should have come back to this months ago.)

core/src/main/scala/spire/math/ULong.scala
```diff
@@ -64,9 +64,13 @@ class ULong(val signed: Long) extends AnyVal {
     if (signed >= 0L) signed.toString
     else (ULong.LimitAsBigInt + signed).toString
 
+  // TODO: replace these with `===` and `=!=`? Don't we get `==` and `!=` for free with value classes?
```
kevinmeredith: From this REPL test (`scala 2.11.7`), it **inconclusively** seems to me that the *free* `==` and `!=` accept an argument of `Any`, not `ULong`, as you wrote in lines 68-69.

I admit that I know nothing about value classes other than reading this article - http://docs.scala-lang.org/overviews/core/value-classes.html.

```
scala> class BoolWrapper(val underlying: Boolean) extends AnyVal
defined class BoolWrapper

scala> class Wrapper(val underlying: Int) extends AnyVal
defined class Wrapper

scala> new Wrapper(5) == new BoolWrapper(true)
<console>:15: warning: Wrapper and BoolWrapper are unrelated: they will never compare equal
       new Wrapper(5) == new BoolWrapper(true)
                      ^
res6: Boolean = false

scala> new Wrapper(5) != new BoolWrapper(true)
<console>:15: warning: Wrapper and BoolWrapper are unrelated: they will always compare unequal
       new Wrapper(5) != new BoolWrapper(true)
                      ^
res7: Boolean = true
```

Also, I read this som-snytt [answer](http://stackoverflow.com/a/18585096/409976) on `why-operator-and-equals-behave-differently-for-values-of-anyval-in-scala`. However, I don't entirely understand it.

kevinmeredith:
Thank you, Erik. Please merge it. 

Once this merges, would it be useful for me to work on https://github.com/non/spire/issues/405 and then https://github.com/non/spire/issues/444?

core/src/main/scala/spire/math/ULong.scala
non: So, you can definitely pass anything to `==` due to Scala's support for Java's universal `.equals` method. The issue about defining the method like this has to do with the bytecode produced.

Here's a gist that look at `==` between value classes versus other types: https://gist.github.com/non/aacc4a86ec5971faea7c

As you can see, the `UByte == UByte` case is handled with a special method (located in the `UByte companion`) called `$eq$eq$extension`. Our method definition is "overriding" that one (maybe unnecessarily).

We do get these for free, so I think they can be removed, but I just wanted to point out the machinery being used.

non:
@kevinmeredith Yes, those make sense. You may want to wait a day or two -- there will be a big PR that moves tons of source files to add scala.js support. I think that will create a lot of conflicts with existing branches, unfortunately.

PR #477, opened 2015-07-24
Comment out rational root finder test

rklaehn:
it causes frequent build failures. I will open an issue
to make sure we don't forget about it.


non:
:+1:

PR #480, opened 2015-07-24
Coverage testing and publishing to codecov.io

rklaehn:
This adds coverage testing to travis ci, and publishes the results to codecov.io.

Coverage testing is for now only run for 2.11. It might be possible to also run it for 2.10 now that I have excluded spire.macros.

@non you will have to configure your codecov.io to exclude some directories using "integration and setup/ignore files". Here is my ignore list:

```
laws/**
benchmark/**
examples/**
macros/**
_2.11/**
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,12 @@ scala:
   - 2.10.2
   - 2.11.6
 script:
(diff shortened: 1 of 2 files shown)


non:
:+1:

PR #485, opened 2015-07-26
Add scala-js support

InTheNow:
This is not yet complete - but should contain everything other than the actual compilation of scala-js.

Code starting at [L185](https://github.com/InTheNow/spire/blob/topic/sjs/build.sbt#L185) perhaps shows  the most important changes.


macros/src/main/scala/spire/macros/Syntax.scala
```diff
@@ -42,7 +42,7 @@ object Ops extends machinist.Ops {
       (uesc('⊼'), "nand"),
       (uesc('⊽'), "nor"))
 
-  def unopWithEv2[Ev1, R](c: Context)(ev1: c.Expr[Ev1]): c.Expr[R] = {
+  override def unopWithEv2[Ev1, R](c: Context)(ev1: c.Expr[Ev1]): c.Expr[R] = {
```
non: We can remove this method now that we are using a more recent version of Machinist.

non:
Other than one comment, :+1: from me.

non:
The 2.11 Travis failure is the well-known algebraic root-finding issue, which we can ignore for now.

The 2.10 Travis failure is a consequence of how slowly Spire compiles under 2.10.5. We may need to disable Travis for 2.10, or figure out how to speed up the build.

non:
(I'm assuming we have to upgrade to 2.10.5 for scala.js? Is it possible to stick with Scala 2.10.2? If so, that would be helpful in terms of Travis/build time. But I am OK with prioritizing scala.js support over Travis 2.10 support if it comes to it.)

non:
I'm going to wait for @tixxit and/or @rklaehn but this looks great to me.

Let's merge it ASAP and then re-target our other work.

InTheNow:
Some general comments on the PR:

* Running the tests found a bug in scala-js `BigInteger` - fixed already https://github.com/scala-js/scala-js/pull/1837 but until released in 0.6.5, all tests cannot pass.
* I added two `Platform` classes [jvm](https://github.com/InTheNow/spire/blob/topic/sjs/core/jvm/src/main/scala/spire/Platform.scala) and [js](https://github.com/InTheNow/spire/blob/topic/sjs/core/js/src/main/scala/spire/Platform.scala) so `TrieMap` is still used in JVM. (And just realised they should have been traits...Shall I change?)
* For the change in `Complex`, see [this](https://gitter.im/scala-js/scala-js?at=555c6668bb71b5b229866cb4)  and [that](https://gitter.im/non/spire?at=555c6dcec88eb7635c269f2e) discussion- I'll summarise the important bits at the end.
* The biggest "pain" was the compile time on travis that just timed out, so I had to serialize the build via the `validate` alias and remove the test compilation as `fastOptJS` just hung. Not sure of the longer term solution here - may need some help from the Scala-js crew on this one. 

Also note that you now have coverage on 2.10 as well as 2.11 (but not for sjs - see https://github.com/scoverage/scalac-scoverage-plugin/pull/118)

The proposed plan is thus to merge this in (pending review, etc) and then add/fix tests and performance problems in parallel to mainstream JVM development.

Summary of chats:

>floatToRawIntBits cannot be implemented, AFAICT. The best that can be done is to implement it as a forwarder to floatToIntBits.But that would not respect its spec, so I would advise against implementing it at all. 
   
>It seems to me they could use floatToIntBits without breaking their use case.
Since they use it to encode Floats into Ints (not encode Ints into Floats).
It is always the case that intBitsToFloats(floatToIntBits(f)) equals f.
What is not always true is that floatToIntBits(intBitsToFloat(i)) equals i. For that you need floatToRawIntBits. AFAICT, they use only the former identity, not the latter.

> Btw, note that the whole point of FastComplex is lost on Scala.js, because a) Longs are excruciatingly slow and b) intBitsToFloat/floatToIntBits are not really fast either (and excruciatingly slow if Typed Arrays are not supported).
FloatComplex would be much faster if it were simply an alias to Complex[Float] on that platform.

non:
It might be good to consider just leaving `FastComplex` out of scala.js entirely. I'm not sure an alias is even necessary -- people should just use `Complex[Float]` explicitly. How many other places do we need to use `floatToIntBits`?

InTheNow:
It's just used there. The advantage of the alias is that code written for JVM (via `FastComplex`) would just work as-is when compiled with scalajs

non:
I'm fine with an alias in principle but I don't think `FastComplex` and `Complex[Float]` have identical APIs, so there still could be source incompatibility issues.

InTheNow:
In that case I suggest we leave as-is for this PR and raise a new issue for this point? (main reason I included it was so that it was not forgotten)

InTheNow:
Or... we could add `floatToRawIntBits` as a Package function ?

non:
Let's keep your solution for now. It won't affect existing users and we can get feedback from folks about what they'd prefer.

build.sbt
```diff
@@ -13,131 +12,191 @@ lazy val spire = project.in(file("."))
...
+lazy val tests = crossProject.crossType(CrossType.Pure)
   .settings(moduleName := "spire-tests")
-  .settings(spireSettings)
-  .settings(testsSettings)
-  .settings(noPublishSettings)
+  .settings(spireSettings:_*)
+  .settings(scalaTestSettings:_*)
+  .settings(noPublishSettings:_*)
+  .jvmSettings(scalacOptions += "-optimize")
+  .jsSettings(testOptions in Test := Seq(Tests.Filter(s => jsTests.contains(s))))
```
tixxit: Would it be OK to invert the logic here and use `jsTests` to filter the tests out. This way new tests are included by default and only excluded when we know they're bad.

InTheNow: I can do, but the biggest issue is with tests that do not return - so you don't actually know what is happening. So the idea here is that I can test them one by one (yawn). Then, it makes sense to filter out the bad ones, sure

core/js/src/main/scala/spire/Platform.scala
```diff
@@ -0,0 +1,11 @@
+package spire
+
+import  scala.collection._
+
+class Platform {
+  type TrieMap[K, V] = mutable.Map[K, V]
```
tixxit: OOC, why put this in `class Platform` rather than putting it in `object Platform` with the `def`?

core/shared/src/main/scala/spire/math/Algebraic.scala
```diff
@@ -591,8 +591,7 @@ object Algebraic extends AlgebraicInstances {
      */
     def flags: Flags = new Flags(flagBits)
 
-    private val bounds: TrieMap[ZeroBoundFunction, Any] =
-      new TrieMap
+    private val bounds: Platform#TrieMap[ZeroBoundFunction, Any] = Platform.TrieMap()
```
tixxit: This is a great compromise - thanks!

tixxit:
This looks great! :+1:

core/js/src/main/scala/spire/Platform.scala
InTheNow: Because .... I had some other stuff in the Class, but not any more. Makes sense to clean it up

non:
Great. Sounds like once we change `Platform` to an object this is good to go.

rklaehn:
LGTM. So after this is merged, Travis will also run scala.js tests and take much longer? So if I just want to do some development locally I should just use validateJVM so I don't have to wait too long, right?

non:
@rklaehn Yes, I think that is correct.

@InTheNow What do you think the best workflow is for someone who wants to do the fastest edit/compile/test cycles on the JVM?

InTheNow:
OK - just finishing of some work that will enable sjs tests, so all issues in this PR are fully covered.

>  I should just use validateJVM so I don't have to wait too long, right?

Absolutely. I will also split the `spire` project into three so `spire` just depends on new `spireJS` and `spireJVM` - this is also useful for publishing, btw.

When developing, say on core, I would tend to just work with `coreJVM/compile` and the relevant `testsJVM/testOnly ...` and occasionally test everything. Scala-js has an incremental linker so after the first run things aren't really that much slower in any case

InTheNow:
Done - first test also running for scala-js.

Build times both ~26min. But this includes jvm running twice, once with coverage instrumentation, once without.

InTheNow:
Also, please note that the sjs build issues were _mainly_ solved by [customising the jvm options for sbt](https://github.com/InTheNow/spire/blob/topic/sjs/.jvmopts). If you have local compile issues, it's probably worth giving these options a try.

non:
:+1:

Merging this.

Thanks for all your help @InTheNow!

Issue #453, opened 2015-05-26
Rational API changes


non:
I'd like to try to improve the `Rational` API a bit. Here are some concrete suggestions:

1. Change `.numerator` and `.denominator` to return `SafeLong`.
2. Add `.isZero` method
3. Consider using `SafeLong` instead of `BigInt` for "big" variant.
4. Add conversion methods to `Real` and `Algebraic`
5. Simplify inheritance scheme if possible.
6. Look into moving `.nroot` to somewhere more appropriate.

(1) is pretty straightforward: creating a `BigInt` from a `Long` is more expensive than wrapping a `Long` or `BigInt`).

(2) is just about creating consistency between our concrete types (and supporting more literate programming: `x.isZero` reads better than `x.signum == 0` for something that is quite common).

(3) is potentially contentious and will require a lot of benchmarking. The thought is that operations on `SafeLong` are not much slower than `BigInt`, will automatically convert to `Long` where possible, and will allow *partial* big rationals (i.e. large numerator with `1L` denominator, or vice-versa). The concern is that extra costs will be significant enough to hurt overall performance in "normal" use.

(4) Is pretty safe, and just about consistency.

(5) Right now there is a tangled web of classes that make up the various rational types. It would be nice to move to something much simpler if possible. However, we have to be sure that any change doesn't negatively impact performance.

(6) We have `.sqrt` and `.nroot` methods for `Int`, `Long`, `BigInt`, and `SafeLong` scattered around Spire. These methods are inexact (and maybe should be renamed) but end up being useful in a lot of situations. Let's try to figure out a strategy that makes these available in an intuitive way.

non:
I'd appreciate any input from folks who've used `Rational`, but especially @tixxit and @rklaehn.

rklaehn:
I very much agree with most points.

(1) Good idea. SafeLong is the spire way to represent an arbitrary size integer, so it is natural to use that even disregarding performance considerations.

(2) I agree completely. In fact I have a isZero enhancement method in my package object for packages that heavily use rationals for this purpose. I also have `def isOne: Boolean = underlying.compareToOne == 0` as a performance optimization for comparing to one, but that might be too specific.

(4) I am not sure if conversions to unrelated types should be part of a type, or if they should be added from outside at some point. But for now I agree about adding them.

(5) Also agree. The traits were originally meant to allow code reuse, I guess. But we ended up implementing them differently for LongRational and BigRational anyway for performance reasons. So it should be possible to simplify this.

About (3)

[...]

PR #487, opened 2015-08-05
Simplifies rational inheritance scheme

rklaehn:
This would be one step of #453 . I think it is preferable to merge this before attempting to do the other items of #453 . Especially 3) is pretty big.

I also made the constructors of SafeLongLong and SafeLongBigInt private[spire], since they allow breaking the invariant that SafeLongLong is always used for numbers that fit into a Long.


non:
Looks good to me.

:+1:

rklaehn:
OK, I will give @tixxit some time to review and then merge it.

rklaehn:
@denisrosset does this look OK?

denisrosset:
Perfect, thank you!

:+1:

tixxit:
Post-merge, but :+1: :)

Issue #489, opened 2015-08-06
Prime number inconsistencies


rklaehn:
There are a number of small bugs/inconsistencies in spire.math.prime:

```
scala> prime.nth(0)
3
scala> prime.fill(10)
Array(2,3,5,...)
scala> (0L until 10L).map(prime.nth)
Vector(3,2,3,5,...)
```

I think nth(0) should return 2. Or at least not 3.

PR #490, opened 2015-08-07
Feature/rational arbitrary

### LINKED ISSUE #484: Improve Rational arbitrary
The rational arbitrary currently produces BigRationals by producing an arbitrary double and converting that to a rational. I think this could be improved by producing a few values at the border from LongRational to BigRational (e.g. a LongRational that becomes a BigRational when negated etc.)

rklaehn:
An attempt to improve the rational arbitrary

Also some tests for the more tricky arbitraries such as rational, safeLong
and interval. The tests check that the arbitraries generate a realistic
distribution of the possible kinds of rationals, intervals or safelongs

They are made deterministic by setting the random seed, so there should not
be any additional transient failures.

Fixes #484

This also adds some simple tests for the spire.math.prime namespace. These will have to be adapted once #489 is fixed one way or the other.

### DIFF
--- a/core/shared/src/main/scala/spire/math/prime/Siever.scala
+++ b/core/shared/src/main/scala/spire/math/prime/Siever.scala
@@ -38,7 +38,7 @@ import SieveUtil._
  * will instantiate a Siever for you with reasonable parameters.
  */
 case class Siever(chunkSize: Int, cutoff: SafeLong) {
-  require(chunkSize % 480 != 0, "chunkSize must be a multiple of 480")
+  require(chunkSize % 480 == 0, "chunkSize must be a multiple of 480")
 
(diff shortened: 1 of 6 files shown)


rklaehn:
Now that travis has miraculously run through without any issues, it would be nice go get a +1 from @non, @tixxit or @denisrosset

This contains a few more things than just the arbitrary improvements. It might slow down the tests a bit since the rational arbitrary now generates big rationals a bit more often. But I think this has to be accepted, because an arbitrary that will only rarely produce certain kinds of values just produces a false sense of security...

non:
:+1:

non:
Thanks for this, especially for the extra tests! Merging now.

Issue #491, opened 2015-08-08
BUG: `Real.toRational` hangs on `pow` function 


hmf:
Trying to use the `Real` to calculate an exponential using the latest stable version (0.10.1). When I create the type no problem. Its when the `toRational` is executed, whether directly or not, that it hangs.
Examples (all those in comments will hang):
```scala
    //val r2 = Real(2).fpow(Real(3))
    //val r2 = Real.e.fpow(Real(3))
    //val r2 = Real(2).pow(3.0)
    //val r2 = Real(2.0).pow(3.0)
    //val r2 = Real(2).fpow(3)
    val r2 = Real(2).pow(3)

    val s = r2.toRational // hangs here
 ```

The power function seems to work only when the exponent is an integer. I tracked this down to the class:
```scala
case class Inexact(f: Int => SafeLong) 
```
When the `apply`is invoked it enters the `None` (second) case. Then it execute `f(p)` and `Inexact.apply` is called again. The looping seems to start here.

This is a show stopper for me. Any way I can get around this?

non:
@hmf Ouch! I'll investigate now. Thanks for reporting.

hmf:
@non  I have tried to search a little more and it seems to lead to Rational@90: 
```scala 
def limitToInt: Rational = limitTo(BigInt(Int.MaxValue)) 
```
I now wonder if this is not the same issue as #393

non:
OK, I see what is happening. It is an easy bug to fix.

As a work-around you can break apart a rational number and apply the `.pow` and `.nroot` calls separately. I'll try to get a fix released ASAP.

non:
Yeah, it's definitely related to a problem with `.limitToInt`. In the case we care about it is easy to fix -- I'll have to see if it fixes that bug too.

hmf:
Thanks for the suggested workaround. However I am working with `Real`. So I guess I have to convert to `Rational` first.  If so I think that won't do. The `Real.toRational` is what's getting me into trouble.  :-(

non:
I'll try to get a release out with a fix for this ASAP.

hmf:
Thank you very much.

Issue #393, opened 2015-02-22
Bug in Rational.limitTo


rklaehn:
The following code hangs
```
(Rational(1,Long.MaxValue) + 1).limitToLong
```

limitToLong just calls limitTo(Long.MaxValue).

tixxit:
Is it hung or just very slow? There is a very bad worst case behavior currently and a comment with a suggestion on how to fix it (I hit the same issue recently) :/

rklaehn:
It might be just very slow. I gave up after 5 minutes. But from a user POV it does not really matter if is theoretically terminating.

tixxit:
Oh yeah, I agree. More making sure this is the same bug and not a different one.

kevinmeredith:
I added print statements to try to figure out the slow computation:

```
  def limitTo(max: BigInt): Rational = if (this.signum < 0) {
    -((-this).limitTo(max))
  } else {
    require(max > 0, "Limit must be a positive integer.")
    println("half")
    val half = max >> 1
    println("floor")
    val floor = this.toBigInt
    if (floor >= max) {
      Rational(max)
    } else if (floor >= (max >> 1)) {
      Rational(floor.toLong)
    } else if (this < Rational(1)) {
      limitDenominatorTo(max)
    } else {
      println("final else")
      limitDenominatorTo(max * denominator / numerator)
    }
  }
```

Then I ran the example from the bug report.

**output**

>scala> (Rational(1,Long.MaxValue) + 1).limitToLong
half
floor
final else

But, from the REPL directly, I'm not sure why I can compute the following rather quickly (< 1 second on REPL)

```
scala> (Rational(1,Long.MaxValue) + 1)
res0: spire.math.Rational = 9223372036854775808/9223372036854775807

scala> val max = BigInt(Long.MaxValue)
max: scala.math.BigInt = 9223372036854775807

scala> val denom = res0.denominator
denominator   denominatorAsLong

scala> val denom = res0.denominator
denom: BigInt = 9223372036854775807

scala> val num = res0.numerator
numerator   numeratorAsLong

scala> val num = res0.numerator
num: BigInt = 9223372036854775808

scala> (max * denom) / num
res4: scala.math.BigInt = 9223372036854775806
```

PR #492, opened 2015-08-08
Fix limitToInt to work around commonly-occuring case.

non:
Previously, calling Rational(3).limitToInt would invoke the
general .limitTo code which is already implicated in at
least one other bug (#393). So, this change does a very
simple check to see if limiting is even necessary before
calling into that code.

This fixes the repoted issue, namely the situation where
Rational(3).limitToInt seems to hang.

Tests are added both for the .limitToInt behavior and
also simple tests of Real#fpow.


non:
Fixes #491. /cc @hmf.

non:
The tests locally worked for me. Unfortunately until we get things sorted out with scala.js we may have to disregard Travis on this.

non:
Due to conflicts I'm going to close this issue and open a new one.

Issue #495, opened 2015-08-10
Add typeclass instances for java.math.BigInteger


rklaehn:
That way one can use java.math.BigInteger with operators and without wrapping overhead.

PR #497, opened 2015-08-11
java.math.BigInteger typeclass instances

rklaehn:
Added typeclass instances for java.lang.BigInteger

Also adds tests and an arbitrary/gen. Implements #495


rklaehn:
Review by @non @tixxit

non:
:+1:

Issue #374, opened 2015-02-07
Set up spire-extras subproject


rklaehn:
The purpose of this subproject would be to contain types and data structures that would increase the size of the main spire jar too much, but that would be useful to have as part of the spire project.

It might be a good idea to move some less-frequently used parts of spire that cause disproportionally large increase in jar size to this repository as well.

non:
:+1: Now that 0.10.1 is out I will look into this.

PR #503, opened 2015-08-13
Adds spire-extras project

rklaehn:
Implements #374


non:
:+1: Assuming Travis passes (or if it fails for bogus reasons).

PR #504, opened 2015-08-13
Convert SafeLong to use java.math.BigInteger directly internally

### LINKED ISSUE #500: Add JMH benchmarks for SafeLong
We need a way to validate the changes to SafeLong (rely on java.math.BigInteger, etc) are an actual improvement. It would also be nice to have benchmarks to prevent regressions.

rklaehn:
Also add simple JMH benchmark of SafeLong that checks the different cases

Related to #453 

Closes #500 

I decided to not use BigInteger typeclass instances and spire.syntax in SafeLong. That is why the code is a bit ugly but explicit.


rklaehn:
Now that it has miraculously passed the CI without incident, review by @non, @tixxit or @denisrosset . I would like to have a thorough review, since this is a major change.

As you can see, I have not exposed BigInteger in the public API for now.

rklaehn:
It seems that this test failure is real. I will take a look at it.

non:
:+1: when tests pass (or if they fail but should have passed)

PR #506, opened 2015-08-14
Change Arbitrary[Positive[A]] to do less filtering.

non:
As @rklaehn observes, filtering a significant proportion of
values from a Gen[A] creates a real risk of tests failing
for not finding enough valid values.

In many cases (as here), .filter can be replaced by .map.


non:
I'm merging this, test failure seems bogus.

PR #509, opened 2015-08-15
Fix limitToInt to work around commonly-occuring case.

non:
Previously, calling Rational(3).limitToInt would invoke the
general .limitTo code which is already implicated in at
least one other bug (#393). So, this change does a very
simple check to see if limiting is even necessary before
calling into that code.

This fixes the repoted issue, namely the situation where
Rational(3).limitToInt seems to hang.

Tests are added both for the .limitToInt behavior and
also simple tests of Real#fpow.


non:
superseded by https://github.com/non/spire/pull/697

PR #510, opened 2015-08-16
Revert "Use upper case for constructor methods"

rklaehn:
This reverts commit 4c2570cb14fed885a4c003d44ccadcfaf5d9ed90.

Apparently, having a method with the same name as an inner class causes problems in 2.10. In any case, all of
this is just cosmetic and hidden from the outside, so reverting this is not a big deal.


non:
Thanks for doing this.

PR #511, opened 2015-08-16
Testing improvements

rklaehn:
Increase coverage for Rational and SafeLong

Avoid wherever clauses that cause occasional build failures


rklaehn:
Example of test failure that should be fixed by this commit:
```
[info]  (-inf, a] does not compare to [b, inf) if a >= b *** FAILED *** (54 milliseconds)
[info]   Gave up after 1 successful property evaluation. 10 evaluations were discarded.
```

rklaehn:
Review by @non, @tixxit or @denisrosset

Mostly test changes, so not much to review...

non:
:+1: Great work, thanks!

PR #513, opened 2015-08-17
Remove a few remaining unnecesary usages of BigInt

rklaehn:
Use SafeLong whenever possible


non:
Looks good to me. :+1: 

Waiting to hear from Travis, hopefully we're finally back in a place where it will go green.

Issue #512, opened 2015-08-17
Explicit Syntax for lifting number literals into spire


JackSullivan:
I've noticed that spire can be a bit rough around the edges when it comes to mixing primitive number literals with spire types. I think it would be worthwhile to add an explicit syntax for bring literals into spire. Concretely:
```scala
def foo[N : Field : Trig : NRoot] = {
  val n = (3.5).as[N]
}
```
This would make it clearer which spire methods are dispatching against primitive types, and make it possible to operate with fewer implicits.

non:
:+1:

This syntax should probably live with the other syntax for literal values, and could look something like this:

```scala
implicit class IntAs(val n: Int) extends AnyVal {
  def as[A](implicit ev: Ring[A]): A = ev.fromInt(n)
}
```

(For `Double` literals, we'd want to use `Field[A].fromDouble`.)

non:
If we wanted to get really fancy we could make these into macros to reduce the overhead when literal values are used (especially for zero and one).

PR #515, opened 2015-08-22
Explicit literals

JackSullivan:
#512  Let me know if this is what you had in mind. I've only implemented `double`s and `int`s since those are the 'default' literal types for scala, and it looks like `Field` and `Ring` specializations are built mainly around them. I'm happy to add `long` and `float`, but I'm not sure what the from method would be for long (that doesn't truncate).

### DIFF
--- a/core/shared/src/main/scala/spire/syntax/Literals.scala
+++ b/core/shared/src/main/scala/spire/syntax/Literals.scala
@@ -64,11 +64,11 @@ class EuLiterals(s:StringContext) {
 }
 
 object primitives {
-  implicit class IntAs(val n:Int) extends AnyVal {
+  implicit class IntAs(n:Int) {
     def as[A](implicit ev:Ring[A]):A = macro Macros.intAs[A]
   }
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/spire/syntax/Literals.scala
```diff
@@ -60,3 +62,13 @@ class EuLiterals(s:StringContext) {
   def big():BigInt = macro Macros.euBigInt
   def dec():BigDecimal = macro Macros.euBigDecimal
 }
+
+object primitives {
+  implicit class IntAs(val n:Int) extends AnyVal {
```
non: We don't need `AnyVal` here, since the constructor call will be rewritten -- I would remove it.

non:
Great! This is exactly what I imagined (modulo the unnecessary `AnyVal` which I mentioned).

JackSullivan:
Ok, I've removed the unnecessary `AnyVal`.

non:
:+1: Looks great to me.

Issue #514, opened 2015-08-21
Do not allow numeric coercions or implicit conversions in ===


rklaehn:
Currently, the === and =!= operators defined in spire.syntax.eq sometimes allow both numeric coercions and implicit conversions.

I think that this should be changed. === and =!= should only compile if lhs and rhs are of the same type.

# Primitive comparisons #
When doing numeric comparisons, it is always possible to just use the right type. With the proposed change, e.g. to compare a double to a constant you would have to write `x === 1d` instead of `x === 1`. The former makes the code much more obvious IMHO and does not represent a significant typing overhead.

# Non-primitive comparisons #
When doing comparisons with spire types such as SafeLong and Rational, I think requiring the same type for LHS and RHS is also clearer. Allowing implicit conversions in something as simple as an equality check introduces difficult to find performance problems, and also makes the code less predictable.

Take the following piece of code:
```scala
def isTwo(x: Rational): Boolean =
  x === 2 // this performs an implicit conversion of 2 to Rational using Rational#apply(Int)
def isTwoNewSyntax(x: Rational): Boolean =
  x === Rational(2) // using the new === syntax, you have to convert rhs to the same type of lhs explicitly
```

I think the second piece of code makes it more clear what is happening, and also shows that a small performance gain could be had by moving Rational(2) into a constant. (I discovered and removed quite a few unnecessary performance degradations in Rational and SafeLong that were caused by an innocent-looking equality comparison)

# General considerations #
One very basic definition of equality is that two things are equal if they can be used interchangeably without observable effect. Obviously, this is not the case for the current === implementation since just because e.g. there is a conversion/coercion from Int => Long or Int => Rational, I can not substitute an Int for a Rational and expect the result of my code to be the same.

With the new, more strict definition of === and =!= we would get much closer. And if you don't want this strict definition there is always == and a yet-to-be-defined mechanism for "cooperative" equality.

# Backwards compatibility #

The only downside I can see in this proposal is that it might break code that uses === and relies on the fact that coercions and implicit conversions do apply to the rhs. But the good thing about this compared to changing *anything* about equals/hashcode is that you will get an immediate compile error instead of subtle, undetected breakage. So I think the change is acceptable.

Here is the current behavior, which allows numeric conversions of the rhs but not of the lhs:
```
scala> import spire.math._; import spire.implicits._
import spire.math._
import spire.implicits._

scala> 1L === 1
res8: Boolean = true

scala> Rational(1) === 1
res9: Boolean = true

[...]

PR #518, opened 2015-08-25
Added macros for strict equality and used them from === and =!=

rklaehn:
Added macros for strict equality and used them from === and =!=

Added shapeless dependency and some syntax tests, and adapted some
equality checks to make it compile.

Added JMH based benchmark for primitive equality using == and ===


rklaehn:
Implements #514 . Now that it passed the CI, review by @non or @tixxit

non:
:+1:

It's great that those were the only changes you needed! I guess current Spire code is doing an OK job of avoiding these problems?

rklaehn:
There were some issues with Option/None/Some, but nothing else. Using spire.util.Opt would not be a problem.

Some problems might be hidden by the === and =!= methods on spire data types. In the long term I would like to remove those. The macro === is just as fast, and more typesafe and symmetric.

PR #520, opened 2015-09-25
Jetdim compile time check

non:
This is a port of https://github.com/non/spire/pull/438 which includes an extra commit.

I'd be happy to merge this and then consider the fate of `NonZero`, `NonNegative`, etc.

What do you all think?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -319,7 +319,7 @@ lazy val genProductTypes = TaskKey[Seq[File]]("gen-product-types", "Generates se
 lazy val scalaCheckSettings  = Seq(libraryDependencies += "org.scalacheck" %%% "scalacheck" % "1.12.4" % "test")
 
 lazy val scalaTestSettings = Seq(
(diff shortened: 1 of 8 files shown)


non:
Thanks @kevinmeredith! Good catch. I'll make the constructor `private[spire]`.

kevinmeredith:
Is it worthwhile to just use `Nat` from shapeless? If it's like Idris's `Nat`, then it starts at 0.

However, is there a way to use shapeless Nat's to fail to compile (like this PR's `PositiveInt`) if it equals `Z`, i.e. only compile if it's a `successor`?

denisrosset:
Hi @non, we discussed last year to move `Jet` towards legacy as it is currently not well-defined (is it an associative algebra element? or an automatic differentiation type that pretends to be whatever the underlying scalar is?)

As such, I'd be keen on closing this PR. What do you think?

larsrh:
Closing this because it hasn't seen activity in over a year.

PR #521, opened 2015-09-27
AdditiveMonoid is sufficient for metric space identity law

adampingel:
I have a case where it's not possible to define a Rng for the second type of a MetricSpace, but I can provide an AdditiveMonoid, which is all that's needed to define the identity law.


non:
Makes sense to me. :+1:

adampingel:
The 2.11 build passed, but the 2.10 one failed with:

[error] Could not run test spire.algebra.SignedTest: java.lang.AssertionError: JS test interface sent bad reply: msg:org.scalatest.events.SuiteStarting
org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed

I'm not familiar with this error.  Any advice?

non:
Did the 2.10 tests run for you locally? I can restart it to see if it passes on another try.

non:
Looks like it was a transient failure.

PR #523, opened 2015-10-11
changed DoubleIsReal so that 'ceil' calls ceil instead of floor

bbuchsbaum:


tixxit:
:+1: Should we check the rest of the code base for similar mistakes?

tixxit:
Just, like, `ack --scala ceil.*floor` or something...

adampingel:
The same issue is present in float.

A sequence of github button presses based on this PR resulted in the creation of a "Patch 1" PR: https://github.com/non/spire/pull/524

adampingel:
My search I also found the RoundingMode in BigDecimalRootRefinement's floor method:
https://github.com/non/spire/blob/master/core/shared/src/main/scala/spire/math/poly/BigDecimalRootRefinement.scala#L139  (and also for ceil).

I may not understand the intent, but at first glance they appear to be swapped.

PR #524, opened 2015-10-12
fix DoubleIsReal.ceil and FloatIsReal.ceil

adampingel:


adampingel:
As noted in the previous PR, I'm curious about the rounding mode here: https://github.com/non/spire/blob/master/core/shared/src/main/scala/spire/math/poly/BigDecimalRootRefinement.scala#L139

non:
:+1:

non:
Thanks!

PR #526, opened 2015-10-20
Add IntervalSeq and IntervalTrie to spire-extras

rklaehn:
Adds IntervalSeq and IntervalTrie to spire-extras

IntervalSeq[T] just requires an order instance, whereas IntervalTrie[T] only works for types that can be converted to long while preserving order.

Also moves FixedPoint and FixedScale to spire.math.extras


rklaehn:
This passes all the tests using sbt validateJVM. It also runs through for both JS and JVM and 2.11.

Review by one of @denisrosset, @non, @tixxit

tixxit:
Amazing! I'll start going through it today, but may take a bit, just due to the size.

laws/src/main/scala/spire/laws/arb.scala
```diff
@@ -2,6 +2,8 @@ package spire.laws
 
 import java.math.BigInteger
 
+import spire.math.extras.{FixedPoint, FixedScale}
```
tixxit: I don't think this import is used. There are a few files with just imports added, so we can probably remove the imports for now.

tixxit: Oh wait - I'm guessing `FixedPoint` got moved. Please ignore!

extras/src/main/scala/spire/math/extras/interval/IntervalSeq.scala
```diff
@@ -0,0 +1,687 @@
...
+
+final class IntervalSeq[T] private (
+    val belowAll: Boolean,
+    private val values: Array[T],
+    private val kinds: Array[Byte],
+    private implicit val order: Order[T]) extends IntervalSet[T, IntervalSeq[T]] { lhs =>
+
+  import IntervalSeq._
+
+  private def binarySearch(key: T): Int = {
```
non: Could you have used `spire.math.Searching.search` here?

tixxit: +1

non: (Alternately, if this method performs better than `spire.math.Searching.seach` we should patch that one.)

extras/src/main/scala/spire/math/extras/interval/IntervalSeq.scala
```diff
@@ -0,0 +1,687 @@
...
+      else {
+        // scalastyle:off return
+        return mid
+        // scalastyle:on return
+      }
+    }
+    -(low + 1)
+  }
+
+  def below(index: Int): Boolean = {
```
tixxit: Given that this is the only version that takes an `Int`, is it worth making it (package) private?

extras/src/main/scala/spire/math/extras/interval/IntervalSeq.scala
```diff
@@ -0,0 +1,687 @@
...
+    val index = binarySearch(value)
+    if (index > 0)
+      valueAbove(kinds(index - 1))
+    else if (index == 0)
+      belowAll
+    else
+      below(-index - 1)
+  }
+
+  def apply(value:T) = at(value)
```
tixxit: We should probably add return types to all the public methods here. Most have them, but a few are missing.

extras/src/main/scala/spire/math/extras/interval/IntervalSeq.scala
rklaehn: You're right. I will see if I can use it.

extras/src/main/scala/spire/math/extras/interval/IntervalSeq.scala
```diff
@@ -0,0 +1,687 @@
...
+
+  override def hashCode = {
+    belowAll.## * 41 + Arrays.hashCode(kinds) * 23 + Arrays.hashCode(values.asInstanceOf[Array[AnyRef]])
+  }
+
+  override def equals(rhs:Any) = rhs match {
+    case rhs:IntervalSeq[_] =>
+      lhs.belowAll == rhs.belowAll &&
+        Arrays.equals(lhs.kinds, rhs.kinds) &&
+        Arrays.equals(values.asInstanceOf[Array[AnyRef]], rhs.values.asInstanceOf[Array[AnyRef]])
```
tixxit: I'm kind of surprised that this cast works. An array of primitives should throw a `ClassCastException` when you coerce it to an `Array[Object]`. Does this get elided by the scala compiler or something?

rklaehn: I am always using the AnyRef class tag, even when creating arrays of primitives.

    private def classTag[T] = ClassTag.AnyRef.asInstanceOf[ClassTag[T]]

So everything is always boxed. This might seem inefficient, but if you want an interval set for primitives you should use an IntervalTrie anyway. The alternative would be to require ClassTag all over the place, which is not nice.

tixxit: Ah - awesome. No worries then.

rklaehn: It's a dirty hack, but since I never accept arrays from the outside it should be OK.

rklaehn:
I did the recommended changes. Build still fails with a "JSCom has been closed" error, but works locally using sbt validateJVM

rklaehn:
@non @tixxit since this now passed the CI, it would be good if you could review this. I would like to merge this soon.

non:
:+1:

non:
Thanks so much for this Rüdiger! Really excited.

PR #527, opened 2015-10-21
Remove === and =!= methods

rklaehn:
Remove === and =!= methods, since they are better handled by the === macro now

The only case that is a bit more complex is the Number hierarchy. Tests pass, but I would not be surprised if this was not properly covered. So review by @non @tixxit


denisrosset:
+1, will update your PR if it conflicts with the algebra merge.

Still, you wanted review by @non, @tixxit, so not merging before approval.

denisrosset:
Continued in #619, closing this PR.

PR #530, opened 2015-11-05
Type aliases

paulp:
Allowing `@tailrec`, `@sp[ecialized]`, and `ClassTag` without importing/renaming them into every file.

All approaches were intrusive, so I chose the one which I thought easiest to adapt to future requirements. There's a spire package object containing the new aliases, and all the files which take advantage of them have their package declarations modified. I left the other package declarations as they were, but they can be made consistent with the others if desired.

### DIFF
--- a/benchmark-jmh/src/main/scala/spire/benchmark/jmh/AddBenchmarks.scala
+++ b/benchmark-jmh/src/main/scala/spire/benchmark/jmh/AddBenchmarks.scala
@@ -1,8 +1,8 @@
-package spire.benchmark.jmh
+package spire
+package benchmark.jmh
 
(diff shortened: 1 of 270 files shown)


non:
Looks good to me! :+1:

If you don't mind, let's change all the files. That way we can move code around without running into missing imports.

paulp:
Since the last commit is a bit hidden in the hodgepodge of fonts, boxes, icons, and colors that is github, I will point out that I changed all the files as suggested.

non:
Aha, thanks!

non:
:+1:

rklaehn:
LGTM!

PR #536, opened 2015-12-05
Use typeclasses from algebra

rklaehn:
Delete spire typeclasses and replace them with aliases to algebra

Standard instances are still provided from spire itself for now


tixxit:
This all looks really great! I don't think we should try to incorporate `algebra.std` in this PR though.

rklaehn:
Agree about doing this step by step. Maybe there should be one more pre-algebra release, since this is a major change. Then an initial algebra release, and then incorporation of algebra.std.

denisrosset:
Looks good, great work!

One thing I don't like when moving from Spire to Algebra typeclasses is the usage of `empty` instead of `id`, `remove` instead of `inverse` for monoids/groups.

When doing group theory, `remove` and `isEmpty` do not have the same meaning as `inverse` and `isId`.

What was the rationale behind the non/algebra terminology?

rklaehn:
As a non-mathematician I also find `id` more intuitive than `empty`. Here is the commit that changed it, btw: https://github.com/non/algebra/commit/e0ccca11117187e1fa5914e85f705735ee65fbc2

denisrosset:
I had looked at the commit history in algebra but somehow missed it.

The commit by @non mentions a *compromise*; this is probably because cats uses monoids to merge data structures. The `empty`, `combine` terminology makes sense in that context.

See non/algebra#131

tixxit:
@denisrosset the main motivation was that algebird uses `empty`, `combine`, etc. They have a lot more users than Spire, so it seemed like an acceptable compromise to avoid breaking too much source compatibility.

denisrosset:
@tixxit, thanks for the background; I understand the argument.

Should we then keep `empty` / `combine` / `isEmpty` for the base type class, but add the more standard names (in the mathematical sense) in the spire.syntax package?

I see that there will be no optimal solution there, and you and @non are better placed to decide on a tradeoff.

rklaehn:
I think adding optional standard names might be a good idea. But that is unrelated to this PR, right?

denisrosset:
What is related to this PR is that the methods names of the Semigroup/Monoid/Group hierarchy are changing in a way that is not consistent with the usual mathematical notation.

We first need to acknowledge the change, and then decide on the consequences for Spire.

1) should we argue to name the identity `id` in non/algebra, or add a method alias to the traits ?

2) should we adopt the `empty`, `isEmpty`, `combine` notation, without alternative syntax? In that case, the partial algebra type classes should be made consistent with their total counterparts, including difficult-to-make-sense-of names such as `leftEmpty`, `rightEmpty` instead of `leftId`, `rightId` to describe left and right identities in groupoids.

3) should we adopt the non/algebra base traits, and add syntactic sugar in spire? this is as easy as defining a macro alias for `op` and `isId`, and have implicit classes for Monoid/Group to handle the `.id` method.

rklaehn:
@denisrosset 

I really don't like the new names very much, since they imply some kind of container. But I think at the end of the day it is just names.

For me, the most important thing is to get spire to use algebra as soon as possible, so

- you can use spire and cats in the same project and
- the requirements of spire get considered when algebra evolves

So as far as I am concerned, we should
- do one more non-algebra spire release
- do the "spire goes algebra" release
- start a discussion about the names with everybody (cats, algebird, spire) involved.

I guess only @non can do releases, right?

denisrosset:
@rklaehn 

I'm onboard with your schedule.

Let's move to non/algebra, but introduce syntactic sugar for `id` and `isId` in Spire's `syntax` package. And we don't change the partial algebra stuff.

rklaehn:
I actually like the `neutral` name proposed over in algebra. `leftNeutral` and `rightNeutral` might take some getting used to, but it actually describes what it does pretty accurately...

denisrosset:
Agree 100% with "neutral". It is also used in some mathematical algebra texts, cannot be confused with "user.id" and "Predef.identity", and is as good as "empty" when e.g. talking about list concatenation.

I can start on renaming the partial structures; how? Should we have an algebra integration branch in the main repo?

rklaehn:
If you want, I can just merge my topic/algebra-integration-2 to topic/algebra-integration on non/spire.

rklaehn:
OK, I merged this to non/spire branch topic/algebra-integration-2 (note the -2, since topic/algebra-integration is the one year old branch from @non)

PR #539, opened 2015-12-07
Remove last vestigial code for Rational#nroot.

tixxit:
This code isn't used anywhere in Spire anymore and doesn't make sense in `Rational`. Let's just remove it for now.


rklaehn:
LGTM!

PR #541, opened 2015-12-11
Algebra integration that works with the current algebra

rklaehn:
As requested by @denisrosset


denisrosset:
Great, thanks a lot!

The old discussion is there: non/spire#536

PR #543, opened 2015-12-20
Restrict accessibility of PositiveInt's constructor.

kevinmeredith:
I had been thinking that this PR was fully complete, and that it was simply waiting to be merged. But I was wrong - my mistake for not following through.

But, then I saw these comments https://github.com/non/spire/commit/65f0d4192ab947ae3600bcf3f82798afe65ded57#commitcomment-13447188 and https://github.com/non/spire/pull/520#issuecomment-143431888 to add the `private [spire]` modifier.

As a result of this change, my understanding is that clients/users  of `non/spire` won't be able to use `new PositiveInt` due to the `private [spire]` modifier - which is an improvement. But, any `com.spire` class will be able to bypass the safe constructor, `PositiveInt#apply`, as I understand. 

However, is it possible and worthwhile to add a warning when invoking `new PositiveInt` from **outside of** the `PostiveInt.scala` class? In other words, suppress warnings when calling it from the macro, but warn otherwise.

### DIFF
--- a/macros/src/main/scala/spire/macros/PositiveInt.scala
+++ b/macros/src/main/scala/spire/macros/PositiveInt.scala
@@ -11,7 +11,7 @@ import spire.macros.compat.Context
  * technology requires it to remain public. Instead, you should
  * construct instances via the PositiveInt(_) factory constructor.
  */
-class PositiveInt(val value: Int) extends AnyVal {
+class PositiveInt private [spire] (val value: Int) extends AnyVal {
   override def toString: String = value.toString
 }
 


kevinmeredith:
Could either of you take a look please, @non or @tixxit?

thanks

denisrosset:
> But, any com.spire class will be able to bypass the safe constructor, PositiveInt#apply, as I understand. 

I'd be wary of making Spire too safe, at the expense of performance/flexibility. Here the straightjacket is quite uncontroversial (what's the cost of a `> 0` check?), and I'm not raising any concerns.

Without zero-cost abstractions ala Rust and full dependent types, numerical code like Spire is bound to leave unsafe methods simply for performance reasons.

kevinmeredith:
> I'd be wary of making Spire too safe, at the expense of performance. Here the straightjacket is quite uncontroversial (what's the cost of a > 0 check?), and I'm not raising any concerns.

If I understand correctly, Denis, you're noting that `But, any com.spire class will be able to bypass the safe constructor ...` is OK due to performance reasons?

I'm just making sure that I understood what you said, i.e. whether your comment indicate that a code change should occur in this PR.

denisrosset:
> If I understand correctly, Denis, you're noting that But, any com.spire class will be able to bypass the safe constructor ... is OK due to performance reasons?

OK of course.

I was just pointing that even in Haskell, numerical code has impure aspects like throwing runtime errors (see `hmatrix`).

PR #548, opened 2016-01-11
Deprecate SecureJava.fromBytes

rklaehn:
This would have to be removed completely in one of the next releases. Related to #547 , but not a fix yet.

Review by @non or @tixxit


striation:
:+1:

PR #551, opened 2016-01-13
Override fill methods in LongBasedGenerator to use nextLong

andrebeat:
This PR overrides the fill methods in `LongBasedGenerator` (`fillInts`, `fillShorts` and `fillBytes`) to use `nextLong` and therefore only perform half the calls to the RNG.

### DIFF
--- a/core/shared/src/main/scala/spire/random/Generator.scala
+++ b/core/shared/src/main/scala/spire/random/Generator.scala
@@ -441,7 +441,71 @@ abstract class IntBasedGenerator extends Generator { self =>
 
 abstract class LongBasedGenerator extends Generator { self =>
   def nextInt(): Int =
-    (nextLong >>> 32).toInt
+    (nextLong() >>> 32).toInt
+
+  override def fillInts(arr: Array[Int]): Unit = {
+    var i = 0
+    val len = arr.length
+    val llen = len & 0xfffffffe
+    while (i < llen) {
+      val n = nextLong()
+      arr(i) = (n & 0xffffffff).toInt
+      arr(i + 1) = ((n >>> 32) & 0xffffffff).toInt
+      i += 2
+    }
+
+    if (len != llen) arr(i) = nextInt()
+  }
+
+  override def fillShorts(arr: Array[Short]): Unit = {
+    var i = 0
+    val len = arr.length
+    val llen = len & 0xfffffffc
+    while (i < llen) {
+      val n = nextLong()
+      arr(i) = (n & 0xffff).toShort
+      arr(i + 1) = ((n >>> 16) & 0xffff).toShort
+      arr(i + 2) = ((n >>> 32) & 0xffff).toShort
+      arr(i + 3) = ((n >>> 48) & 0xffff).toShort
+      i += 4
+    }
+
+    if (i < len) {
+      var n = nextLong()
+      while (i < len) {
+        arr(i) = (n & 0xffff).toShort
+        n = n >>> 16
+        i += 1
+      }
+    }
+  }
+
+  override def fillBytes(arr: Array[Byte]): Unit = {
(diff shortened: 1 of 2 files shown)


andrebeat:
I've used the `RandomBenchmark` to test this change on Lcg64.

```
[info]           benchmark    ms linear runtime
[info]       FillIntsLcg64 18.94 ==============================
[info]   FillIntsLcg64Fast  9.38 ==============
[info]     FillShortsLcg64  8.87 ==============
[info] FillShortsLcg64Fast  7.61 ============
[info]      FillBytesLcg64  5.98 =========
[info]  FillBytesLcg64Fast  7.58 ============
[info]
```

The results for `fillBytes` are interesting.

rklaehn:
I reran the tests, and the MarsagliaGaussian is normal test fails consistently. @andrebeat can you check if this is related?

andrebeat:
Indeed the test fails deterministically. The issue is related with the override of the `fillBytes` method but I can't seem to understand what the problem is. I'd like another pair of eyes to look at the `fillBytes` method to make sure I'm not doing something wrong. 

Also, I traced the usage of the `fillBytes` method in that test to the `UniformBigInt` implicit defined in `Uniform.scala`. Again, I couldn't quite understand what's wrong with it but I'm not experienced with that code.

non:
@andrebeat I'll take a look.

non:
I figured this one out, and it's probably my fault.

We were using the LCG64 generator to power the gaussian test. For something like `Double` this is fine, but I think `BigDecimal` needed enough entropy that the limitations of LCG64 cause problems. The previous implementation was only using the upper 32-bits, which (I think) have a better distribution than the lower 32. This meant that once we stopped throwing away half the bits the test fails.

In other words, we want to do this:

```scala
  def checkMarsagliaGaussian[A: Field: NRoot: Trig: IsReal: Uniform: ClassTag] = {
    val gen = rng.Cmwc5.fromTime(42L) // replaces rng.Lcg64
    val gaussian = new MarsagliaGaussian[A]
    checkGaussian[A] { (mean, stdDev) =>
      gaussian(mean, stdDev)(gen)
    }
  }
```

Andre, would you mind adding this to your branch?

Also, given that the performance benchmark shows that filling bytes is faster with the "old" code, maybe we should remove this `fillBytes` override anyway?

Thanks for your work on this!

andrebeat:
Nice catch! I actually tried replacing the hardcoded seed (`42L`) with other seeds as I also suspected that for some reason the `Lcg64` was generating less entropy. I had no idea that the upper 32-bits had a better distribution. I've updated the `checkMarsagliaGaussian` as you suggested.

I ran the `fillBytes` benchmark again with all the `LongBasedGenerator` instances:

```
[info]                      benchmark    ms linear runtime
[info]                 FillBytesLcg64  6.08 ======
[info]             FillBytesLcg64Fast  7.88 ========
[info]                 FillBytesCmwc5 11.39 ============
[info]             FillBytesCmwc5Fast 11.85 ============
[info]     FillBytesMersenneTwister64 28.46 ==============================
[info] FillBytesMersenneTwister64Fast 18.20 ===================
[info]
```

I haven't looked at the generated code but I assume the JVM is doing something clever to make it faster in the "old" code. The override only pays off for complex RNGs so I agree with you that its probably not worth it, I'm just a bit afraid that this may lead to some inconsistencies. For example, in the case of `Lcg64` if we remove the `fillBytes` override, generating bytes will lead to a better distribution than generating ints or shorts.

non:
I think the reason is that the "byte logic" is complicated enough (and the LCG64 is fast enough) that just generating more longs is easier for the JIT to optimize. Anyway, why don't we merge this as-is, and follow up with performance stuff later.

:+1: Thanks!

PR #553, opened 2016-01-20
Feature/classdiagrams

rklaehn:
Add class diagrams to Guide.md

This seems pretty useful for explaining the typeclass hierarchy to people. And it is much nicer than having this done as part of the build. The only downside I can see is that there is this gravizo watermark on the images.

Review by @non or @tixxit

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -319,12 +319,12 @@ and multiplicative variants:
   digraph G {
     aize ="4,4";
     node [shape="box"];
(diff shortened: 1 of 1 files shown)


denisrosset:
Hi @rklaehn! We had a pass through the issues/PR with @non during nescala. We are slightly reluctant to introduce a dependency on a non-free service.

What are the alternatives? Can we introduce a SBT task that delegates to graphviz?

PR #557, opened 2016-02-25
Potential fix for inliner swalling variable values.

bashimao:
Whoops.. I meant to write "swallowing" ;-)

 In response to https://github.com/non/spire/issues/556.

As said in the issue report. My abilities considering debugging Scala macros are limited. I did not really check whether this works for complex macros. It might. But you probably want to confirm that yourself. This can be considered as a suggestion regarding how the bug can be fixed.


non:
Thanks for this! 👍 

I think your fix looks good. Sorry it took me so long to get to it.

PR #558, opened 2016-03-14
Polynomial[BigDecimal].gcd was blowing up the heap

sammthomson:
Includes a regression test and the fix.
Rescaling should only happen when there's no non-trivial divisor anyway.


non:
Thanks for this! 👍 

I ended up adding a commit to fix that test that was broken on 2.10.

PR #560, opened 2016-03-15
Multivariate Polynomials - Update 2016

zmre:
Setting this up as a pull request to master instead of topic/multivariate because my git foo wasn't good enough to get that branch up to snuff.  Here's a summary of the changes:
- Added string parser for monomial
- Overhauled toString
- Added partial evaluator
- Made coefficients without vars use an empty var map
- Switched to SortedMap from Map
- Removed monomial +/- operators in favor of add/subtract so an implicit
  conversion to MultivariatePolynomial is a more likely outcome of the
  addition operator (and a safer likely result)
- Added more test coverage
- Added some info to the contributing doc
- Added an .sbtopts to increase memory limits automatically

All tests are passing.  Let me know what you think.


denisrosset:
Wow. Thanks for the work of reviving this branch!

Let's review the design choices (some come from the original branch):

1. for univariate polynomials, the variable name does not matter. In multivariate, it does. The choice here is to have variables be `Char`.

2. a monomial is an object with a scalar and a `SortedMap` of `Char -> Int`

3. the polynomial is stored as an array of sorted monomials.

There are lots of possible choices here, some of them are only not possible in a managed VM, see [the maple encoding for example](http://www.cecm.sfu.ca/CAG/papers/ascm14.pdf).

First, some comments on the design.

The choices made do not lead to a particularly fast implementation, especially `SortedMap` and the usage of FP methods (for example, the `.tail` call in the `Eq` instance for monomials). That's completely fine for me; if we go that route, we can make the rest type-safe, by wrapping the exposed `Array` in `MultivariatePolynomial`. We could also allow multi-letter variable names (using either `String` or `Symbol`), but this complicates parsing. 

All in all, this would provide a default implementation of multivariate polynomials in Spire, which is a good thing. Maybe because the design choices here are opinionated, move these in `spire/extras`.

An alternative is to focus on performance. In that case, I can propose slight modifications to the design, based on my experience of CAS such as Macaulay2 (as a user).

core/shared/src/main/scala/spire/math/poly/Monomial.scala
```diff
@@ -0,0 +1,375 @@
...
+import spire.implicits._
+import spire.math._
+import scala.util.matching.Regex
+
+
+/**
+ *  A monomial is the product of a coefficient and a list of
+ *  variables (Char as symbol) each to an integer power.
+*/
+class Monomial[@spec(Double) C: ClassTag: Order] private[spire] (val coeff: C, val vars: SortedMap[Char, Int])
```
denisrosset: Why the `ClassTag` ? is it only for the Java equals implementation ?

core/shared/src/main/scala/spire/math/poly/Monomial.scala
```diff
@@ -0,0 +1,375 @@
...
+  implicit def monomialEq[@spec(Double) C: ClassTag: Order: Semiring] = new MonomialEq[C] {
+    val scalar = Semiring[C]
+    val ct = implicitly[ClassTag[C]]
+  }
+
+}
+
+// An equivalent monomial has the same variables (that's all!)
+// not checking that the variable exponents are equal using this instance
+trait MonomialEq[@spec(Double) C] extends Eq[Monomial[C]] {
```
denisrosset: What's the motivation of an `Eq` instance that does not obey the `Eq` laws ? If there is no other solution, we have to hide this and make it private[something].

denisrosset:
Proposition: remove the scalar from `Monomial`, and have the polynomial be a `Map` from monomials to coefficients. This solves many comments in the code review.

core/shared/src/main/scala/spire/math/poly/MultivariatePolynomial.scala
```diff
@@ -0,0 +1,214 @@
...
+import scala.collection.mutable.ArrayBuilder
+import scala.collection.mutable.Set
+import scala.reflect._
+import scala.{specialized => spec}
+import spire.algebra._
+import spire.implicits._
+import spire.math._
+
+
+class MultivariatePolynomial[@spec(Double) C: Order] private[spire] (val terms: Array[Monomial[C]])
```
denisrosset: The `terms` mutable array is public; considering the rest of the code is written without much emphasis on performance, we can replace it by a `Seq`, and forget about the `ClassTag`.

Same remark as above regarding the `Order` requirement: this forbids the use of complex coefficients.

core/shared/src/main/scala/spire/std/sortedMap.scala
```diff
@@ -0,0 +1,141 @@
+package spire
```
denisrosset: The `seq` and `map` instances in Spire are for the generic base type. Here we introduce a specialization for `SortedMap`.

What is the future of this design ? move to higher-kinded instances (often conflicts with specialization) ? add special cases as needed ?

tests/src/test/scala/spire/math/MultivariatePolynomialTest.scala
```diff
@@ -0,0 +1,341 @@
+package spire.math
```
denisrosset: The test case looks good. Property-based testing, and a few hand-written cases that show usage examples.

zmre:
With regards to the design choices:

> for univariate polynomials, the variable name does not matter. In multivariate, it does. The choice here is to have variables be Char.

I think this is a reasonable choice.  I can perhaps imagine someone wanting to use "alpha" or "delta" or something, but I would rather that be transformed into the appropriate UTF8 character anyway.  I have a separate little library that converts a subset of ASCII Math (asciimath.org) syntax into UTF8 and back.  It depends on Scala's parser combinator library, but if the spire group is interested, I could submit that to extras.  That would allow for some specific greek letters and other symbols to be used while keeping the variable as Char.

> a monomial is an object with a scalar and a SortedMap of Char -> Int

In a different comment you proposed separating out the scalar so that a monomial was only some mapping of Char -> Int.  I am against that proposal.  You cannot combine two x's via addition without a coefficient.  To separate it out would likely make the maintenance of the feature a major pain.

Regarding the choice of SortedMap vs. some other choice such as the mentioned Maple Encoding: I'm largely impartial.  It is a private implementation detail so the tradeoffs are speed and convenience.  The SortedMap is quite convenient and easy to work with.

> the polynomial is stored as an array of sorted monomials.

Again, this could be accomplished with a specialized implementation.  This implementation is pretty easy to understand and work with as a developer, but is likely less efficient than a more specialized implementation.  

> we can make the rest type-safe, by changing the exposed Array in MultivariatePolynomial to a Seq

Are you referring to the `allTerms` method?  That's an easy change to make.  We could similarly change the `allVariables` method.  The internal `terms` property is not exposed directly.

> We could also allow multi-letter variable names (using either String or Symbol), but this complicates parsing

Covered already, but I would argue against this with the caveat of maybe allowing some special substitutions for greek letters and such.

> Maybe because the design choices here are opinionated, move these in spire/extras.

Wherever you want to put it is fine with me, but I'd argue that the public interfaces are ignorant of the underlying design choices.  It could certainly be evolved with some more efficient encodings.

> Proposition: remove the scalar from Monomial, and have the polynomial be a Map from monomials to coefficients. This solves many comments in the code review

I interpret this as moving the coefficient up to the polynomial.  You'd probably want a sorted map so you didn't have to sort on every operation that returns the data.  You could do this, but you lose the nice ability of letting monomial handle the math.  In fact, you'd probably remove much of the code from monomial into polynomial.  This would make the polynomial class quite a bit bigger and I'm not sure I understand what would be gained except the ability to quickly find like monomials, which would be useful to be sure.  Looking back at the history of the original code though, it looks like the original contributor had a `Map` to begin with and changed it to `Array` after some discussion here on github.  Overall, I'm open to the change, but I don't see a compelling reason for it at the moment.

On the design goals in general: my personal needs here are to be able to more generically model some specific math.  Most of what I'm doing is not performance sensitive.  Two examples are a way to generically work with [division polynomials](https://en.wikipedia.org/wiki/Division_polynomials) where the A and B parameters come in later, and a way to model and work with [changes of variables](https://en.wikipedia.org/wiki/Change_of_variables).  My use cases involve finite fields and an implementation of them that works well with Spire so I can more readily do math over polynomials on extension fields, etc. 

> are the variables assumed to be commutative (i.e. the monomial x * y and y * x are the same) ? this is fine, but it should be documented

Yes, that is assumed.  Where is the best place to document it?  GUIDE.md?  In the code?

core/shared/src/main/scala/spire/math/poly/Monomial.scala
zmre: I'm not sure on either question (`ClassTag` or `Order`).  They were in the original PR.  I can't imagine why Monomial would need to order anything based on coefficient.  Let me see if I can remove both of them.  If I can't, I'll explain why not.

core/shared/src/main/scala/spire/math/poly/Monomial.scala
zmre: This is silly.  I didn't look closely at this, but you're right, it should not treat two Monomials as equal.  Anything wanting to see if the vars are equal should check that instead.  I'm starting to see why you wanted to move the coefficient out.  For now, I'll fix this by making the `MonomialEq` work as you would expect and fixing anything that relied on the broken behavior.

core/shared/src/main/scala/spire/math/poly/MultivariatePolynomial.scala
zmre: I'll try to get rid of `Order`.  Also, pardon my ignorance, I had thought that making the constructor private also made the class members in the constructor private.  I'll go through and make `terms` private here and do the same thing for Monomial's class members and make sure there are methods to access them.  These should be hidden implementation details.

core/shared/src/main/scala/spire/std/sortedMap.scala
zmre: Making this a bit uglier: the implementation is basically identical as for Map.  Is there a pattern elsewhere in spire that should be followed here?

non:
Hi @zmre, thanks for picking this up. (And thanks to @denisrosset for reviewing.)

Since no one has had specific multivariate requirements, I'm happy to defer to @zmre's use case (i.e. a convenient implementation without a focus on performance) for now. I agree with @denisrosset's comments about removing some of the cruft around `Array`, `ClassTag`, and such. (I probably counseled in favor of those, but they do complicate the implemenetation.)

The guide is the best place (for now) to put usage notes, requirements, and limitations, such as the commutativity requirement for *xy* versus *yx*, although it is worth briefly mentioning in the code as well.

denisrosset:
Hi @zmre,

thanks for the detailed answer. When listing the design choices, my intention was not to criticize them, but to make a summary for the other Spire maintainers.

1. I'm fine with the use of `Char`; can you add to the documentation the allowed characters (from the regex  `val variable = "[\u03B1-\u03C9a-zA-Z]".r`) ?

2. Regarding `ClassTag`, I guess if we want specialization and fast sorting we have no choice.

3. I thought again about the MultivariatePolynomial / Monomial split. There are actually two definitions of monomials, says [Wikipedia](https://en.wikipedia.org/wiki/Monomial).

The structure will be clearer if we define a monomial to be a product of powers of variables.

- A monomial is a product of variables raised to some positive power. An empty monomial is just the unit of the underlying Ring. Monomials form a multiplicative monoid, with addition of powers as the product, and the empty monomial as `one`.

- Total orders are defined on monomials without any requirement on the underlying ring/field. Thus we can keep the lexicographic, graded, ... orderings whatever the underlying ring/field is.

- A multivariate polynomial is a vector space (for a scalar field), or a free module (for a scalar ring) over the monomials. 

The current definition (monomial = coeff + variables) is less suitable, and this manifests in requiring the ring/field to be ordered, ruling out the usage of polynomials with complex numbers, or finite fields (that cannot be ordered). And the `add` method on monomials is unsafe.

I see a way out, let me try it.

denisrosset:
I added tests for the monomial ordering in my version --- it seems that the reverse of the sorted map here does not do anything in the graded reverse lexicographic ordering here.

denisrosset:
I made a version with the discussed split, see #562

Currently, a work in progress. See:

https://github.com/denisrosset/spire/blob/multivariate-update-2016/core/shared/src/main/scala/spire/math/poly/Monomial.scala

and 

https://github.com/denisrosset/spire/blob/multivariate-update-2016/core/shared/src/main/scala/spire/math/poly/MultivariatePolynomial.scala

Major changes:

- I dropped SortedMap, and store everything in Array pairs; however, I only expose the arrays in the [spire] context,
- I store the hashCode and degree of monomials in lazy vals to speed-up the monomial orders (used a lot!),
- dropped a lot of the companion construct methods, some of them were redundant. Let's add what we need in a clean way,
- added lots of tests for the monomial and monomial orderings; I'm not convinced the original code is bug-free, I caught several mistakes in mine (orderings are tricky to get right),
- dropped support for negative exponents (they are not defined mathematically, and do not play well with orderings),
- renamed gcd to gcf (happy to revert this),
- used `Opt[Monomial]` as a result for the division,
- I require `CRing[C]` for `MultivariatePolynomial`; do other algebraic structures make sense ?
- added a method to change the ordering of the polynomial
- dropped the euclidean division methods on MultivariatePolynomial; they depend on a particular ordering, and we do not encode the `Order[Monomial]` in the polynomial type (we could! using singleton types, but I feel this is contrary to Spire design goals). I plan to reintroduce them in the `EuclideanRing` typeclass, which will be instantiated with an explicit monomial order.

MultivariatePolynomial is WIP, and the tests have not yet been restored (renamed to .old for now).

zmre:
I'll take a look at the code later today, but wanted to respond to your comments about it.  Everything sounds good with this one exception: 

> dropped support for negative exponents

In the simple case of dividing `x` by `z`, the result is `x/z` which would be represented as `xz^-1`.  If you don't have negative exponents, you really can't have a division operation.  Without division, this class becomes useless for my needs.  Please do find a way to allow negative exponents.

non:
(Disclaimer: I am not an expert here.)

When I was looking into this issue, I came across this useful answer on MathExchange: http://math.stackexchange.com/questions/574001/why-cant-polynomials-have-negative-exponents-or-division-by-a-variable.

According to that answer, there are three types of polynomials we could choose to implement:

 * no negative exponents (no division)
 * ring of rational functions (closed under division)
 * laurent polynomials (allowing negative exponents, division may be complicated [1])

Since @zmre has a use case I think we should be sure to support it. However, since negative exponents aren't part of the "standard" polynomial definition I wonder if we should rename this class? I'd like to be able to point to a standard definition of some kind to explain what properties the type is expected to have.

[1] http://math.stackexchange.com/questions/125573/how-do-i-divide-laurent-polynomials

non:
For example, Wikipedia defines the division polynomials as a sequence of polynomials in Z, but the linked Wikipedia definition for polynomials says that they only employ "non-negative integer exponents". My guess is that division polynomials are defined using Laurent polynomials? @zmre do you have a sense of what is going on with these definitions?

zmre:
@non, you are correct (and I stand corrected) that polynomials by definition have positive exponents.  It does seem that Laurent Polynomial is the best fit for an equation with terms that may have negative exponents.

I *think* regular polynomial long division works with laurent polynomials, but will do some research to confirm that.  If we should want to do a multivariate implementation that is positive exponents only, I expect division will need to be taken off the table or at least reworked.  The current division implementation will return negative exponents.

denisrosset:
My main problem with negative exponents is that I do not know a correct definition of monomial orderings on them (lex, grad lex, grad rev lex). The resources I have only deal with standard polynomial with positive exponents.

There are other things to solve as well:

- you can have monomials of degree 0 that are not the unit, i.e. `x z^-1` is of degree 0 but is not equal to 1,

- to evaluate a monomial with negative exponents, you need a `Field[C]`, otherwise you only need `CRing[C]` --- so some of the methods will be either too restrictive (requiring a field), or not type-safe (requiring a ring but bombing when exponents are negative).

@zmre, would using rational functions cater for your use case ? Once we have standard multivariate polynomials (and a good, safe design for Euclidean division, i.e. a way to specify a default monomial order), they are mathematically sound.

denisrosset:
See https://en.wikipedia.org/wiki/Rational_function

zmre:
Interesting points, @denisrosset.  

Rational function: might work.  Would need to be able to mix and match so you could have `x/y + x/z` and then operate on that (eval, simplify, etc).  Also would need the rational functions to be multivariate.  Would you implement that using the existing `Rational` class?  Can you give me some more hints on what that might look like?

If that doesn't work, then I'm not too worried about requiring `Field` for `eval` and I don't think there are any major consequences to the ordering conundrum. As far as I know, none of the math operations depend on the order.  It does overcomplicate things for the simpler use cases though.

zmre:
@denisrosset - your code looks good to me. It would be nice to bring back string parsing on the polynomial.  I found your tests on Monomial and I should perhaps stand corrected on the ordering.  The reference to the MIT notes shows that there are concepts that depend quite definitely on the order.  

Your approach is much more spire-esque as well (with the arrays and cfors).  I'm happy to defer to it.  Would definitely like to figure out how to get it to a point where more complex equations can be modeled and evaluated.

denisrosset:
Closed during clean-up.

PR #562, opened 2016-03-16
Multivariate update 2016 - WIP

denisrosset:
Work in progress. I made a different split between Monomial / MultivariatePolynomial.


denisrosset:
As noted in #560 

Major changes:

- I dropped SortedMap, and store everything in Array pairs; however, I only expose the arrays in the [spire] context,
- I store the hashCode and degree of monomials in lazy vals to speed-up the monomial orders (used a lot!),
- dropped a lot of the companion construct methods, some of them were redundant. Let's add what we need in a clean way,
- added lots of tests for the monomial and monomial orderings; I'm not convinced the original code is bug-free, I caught several mistakes in mine (orderings are tricky to get right),
- dropped support for negative exponents (they are not defined mathematically, and do not play well with orderings),
- renamed gcd to gcf (happy to revert this),
- used `Opt[Monomial]` as a result for the division,
- I require `CRing[C]` for `MultivariatePolynomial`; do other algebraic structures make sense ?
- added a method to change the ordering of the polynomial
- dropped the euclidean division methods on MultivariatePolynomial; they depend on a particular ordering, and we do not encode the `Order[Monomial]` in the polynomial type (we could! using singleton types, but I feel this is contrary to Spire design goals). I plan to reintroduce them in the `EuclideanRing` typeclass, which will be instantiated with an explicit monomial order.

MultivariatePolynomial is WIP, and the tests have not yet been restored (renamed to .old for now).

denisrosset:
I'm still on this, however I'd like to understand better the gcd, quot/mod stuff in algebra (typelevel/algebra#172) before resuming work.

denisrosset:
Closed during clean-up.

PR #573, opened 2016-06-15
make Interval extend Serializable

scubamaggo:
I did not create any unit tests for serializability as none exist for the other serializable classes. If they would be welcomed I could add some


non:
👍 If you feel like adding tests it's fine, but I'm also fine with merging this sooner.

Let's see what other maintainers think.

rklaehn:
LGTM

Issue #574, opened 2016-06-24
Integral[Int]'s toBigInt method call ends in infinite recursion


pedropramos:
I've noticed an odd case of infinite recursion when calling implicit methods `.factor` and `.isPrime` for instances of `Int`. This can be reproduced with the following interaction:

```
scala> import spire.implicits._
import spire.implicits._

scala> 100 .factor
```

which ends in a stack overflow from mutual recursion, as evidenced by the following stacktrace:

```
...
at spire.math.IntIsIntegral.toBigInt$mcI$sp(Integral.scala:43)
at spire.std.IntIsReal$class.toBigInt(int.scala:65)
at spire.math.IntIsIntegral.toBigInt$mcI$sp(Integral.scala:43)
at spire.std.IntIsReal$class.toBigInt(int.scala:65)
at spire.math.IntIsIntegral.toBigInt$mcI$sp(Integral.scala:43)
at spire.std.IntIsReal$class.toBigInt(int.scala:65)
at spire.math.IntIsIntegral.toBigInt$mcI$sp(Integral.scala:43)
```

As far as I can tell, `IntIsIntegral`'s `toBigInt` method relies on `IntIsReal`'s `toBigInt`, which should be a straightforward call to `BigInt`'s constructor.

Any clue on why `IntIsIntegral`'s `toBigInt` is being called again?

PR #575, opened 2016-06-28
fix infinite recursion on Integral[Int] methods

pedropramos:
solves infinite recursion for methods .factor, .isPrime, and .toSafeLong over Ints
(solves #574)


non:
Thanks! 👍

PR #576, opened 2016-09-13
Update Spire to depend on typelevel/algebra

non:
This is a really big PR. It introduces a dependency on _algebra_ (and transitively on _cats-kernel_) for many of its fundamental type classes. We don't currently have version-to-version compatibility guarantees, but this is still going to be a breaking change for some people.

(Thanks to @rklaehn for taking the first stab at it and getting us most of the way there.)

The most noticeable change is the addition of a `Gcd` type class, and the removal of gcd-related methods from `EuclideanRing`. There are aliases in place so we can still refer to `spire.algebra.Ring`, which is now an alias for `algebra.ring.Ring`. I left all our instances as-is -- we're not currently reusing any instances for _algebra_.

At this point, I have all the tests passing, and have removed all the non-test warnings. Before merging I'd like to do the following:
- [ ] Remove all test warnings
- [ ] Run benchmarks here and in _master_ to ensure no performance regressions
- [ ] Update the documentation to explain where the external type classes are coming from

Other non-mandatory changes that might be nice to have are:
- Update law-checking to use algebra's exported laws.
- Use algebra's instances in some cases.
- Restructure build files to be more similar to these newer projects

We're currently on Spire 0.11.0. I think we should do a release just before merging this PR, to ensure that every other bugfix/feature we have is available before the change. This is definitely one of the last milestones we need before a future Spire 1.0.0 release.

Review by @tixxit, @johnynek, @rklaehn, and others.


tixxit:
OMG, thanks so much for doing this!!!! I'll start plodding through it!

benchmark/src/main/scala/spire/benchmark/Benchmark.scala
```diff
@@ -9,10 +9,6 @@ import spire.math._
 
 import com.google.caliper.Runner
 import com.google.caliper.SimpleBenchmark
-import com.google.caliper.Param
-
-import java.lang.Math
-import java.math.BigInteger
```
tixxit: Thanks for cleaning up all these imports! This was from the warnings?

core/shared/src/main/scala/spire/algebra/Gcd.scala
```diff
@@ -0,0 +1,25 @@
...
+  def lcm(a: A, b: A): A
+
+  @tailrec final def euclid(a: A, b: A)(implicit eq: Eq[A], er: EuclideanRing[A]): A =
+    if (b === er.zero) a else euclid(b, a % b)
+}
+
+object Gcd {
+  @inline final def apply[A](implicit ev: Gcd[A]): Gcd[A] = ev
+
+  implicit def fromEuclideanRing[@sp(Int, Long, Float, Double) A: EuclideanRing: Eq] =
```
tixxit: Let's not make this `implicit`. I'd much rather people explicitly opt-in to this. For example, rational numbers require quite a bit of care, and polynomials need to deal with an edge case or 2.

core/shared/src/main/scala/spire/algebra/Gcd.scala
```diff
@@ -0,0 +1,25 @@
...
+import scala.{specialized => sp}
+
+import spire.syntax.eq._
+import spire.syntax.euclideanRing._
+
+trait Gcd[@sp(Int, Long, Float, Double) A] extends Any {
+  def gcd(a: A, b: A): A
+  def lcm(a: A, b: A): A
+
+  @tailrec final def euclid(a: A, b: A)(implicit eq: Eq[A], er: EuclideanRing[A]): A =
```
tixxit: We can probably move this to the companion object, since it doesn't actually use `gcd` or `lcm`.

core/shared/src/main/scala/spire/algebra/package.scala
```diff
@@ -0,0 +1,162 @@
...
+  type AdditiveCMonoid[A] = _root_.algebra.ring.AdditiveCommutativeMonoid[A]
+  val AdditiveCMonoid = _root_.algebra.ring.AdditiveCommutativeMonoid
+
+  type AdditiveGroup[A] = _root_.algebra.ring.AdditiveGroup[A]
+  val AdditiveGroup = _root_.algebra.ring.AdditiveGroup
+
+  type AdditiveAbGroup[A] = _root_.algebra.ring.AdditiveCommutativeGroup[A]
+  val AdditiveAbGroup = _root_.algebra.ring.AdditiveCommutativeGroup
+
+  object Additive {
```
tixxit: I don't have strong opinions either way, but if you wanted to clean this up, `Additive` and `Multiplicative` can stay in its own file.

core/shared/src/main/scala/spire/algebra/package.scala
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    def apply[A](m: CMonoid[A]): MultiplicativeCMonoid[A] = new MultiplicativeCMonoid[A] {
+      def times(x: A, y: A): A = m.combine(x, y)
+      def one = m.empty
+    }
+
+    def apply[A](g: Group[A]): MultiplicativeGroup[A] = new MultiplicativeGroup[A] {
+      def times(x: A, y: A): A = g.combine(x, y)
+      def div(x: A, y: A): A = g.combine(x, g.inverse(y))
```
tixxit: We should use `g.remove(x, y)` here instead, in case it's been overidden.

core/shared/src/main/scala/spire/algebra/package.scala
```diff
@@ -0,0 +1,162 @@
...
+      def one: A = g.empty
+      override def reciprocal(x: A): A = g.inverse(x)
+    }
+
+    def apply[A](g: AbGroup[A]): MultiplicativeAbGroup[A] = new MultiplicativeAbGroup[A] {
+      def times(x: A, y: A): A = g.combine(x, y)
+      def div(x: A, y: A): A = g.combine(x, g.inverse(y))
```
tixxit: Same as above, `g.remove(x, y)`

core/shared/src/main/scala/spire/macros/Auto.scala
```diff
@@ -200,15 +198,18 @@ abstract class AutoAlgebra extends AutoOps { ops =>
...
-  def times[A: c.WeakTypeTag]: c.Expr[A] = binop[A]("$times")
-  def negate[A: c.WeakTypeTag]: c.Expr[A] = unop[A]("unary_$minus")
-  def quot[A: c.WeakTypeTag]: c.Expr[A] = binopSearch[A]("quot" :: "$div" :: Nil) getOrElse failedSearch("quot", "/~")
-  def div[A: c.WeakTypeTag]: c.Expr[A] = binop[A]("$div")
-  def mod[A: c.WeakTypeTag](stub: => c.Expr[A]): c.Expr[A] = binop[A]("$percent")
-  def equals: c.Expr[Boolean] = binop[Boolean]("$eq$eq")
+  // we munge these names with dollar signs in them to avoid getting
+  // warnings about possible interpolations. these are not intended to
+  // be interpolations.
+  def plusplus[A]: c.Expr[A] = binop[A]("$" + "plus" + "$" + "plus")
```
tixxit: OMG - is this to fix warnings? lol :facepalm:

core/shared/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -490,14 +485,16 @@ with EuclideanRing[Polynomial[C]] with VectorSpace[Polynomial[C], C] {
...
+    val result = euclid(x, y)(Polynomial.eq, this)
     if (result.degree > 0) {
       result
     } else {
       // return the gcd of all coefficients when there is no higher degree divisor
       Polynomial.constant(spire.math.gcd(x.coeffsArray ++ y.coeffsArray))
     }
   }
+
+  def lcm(x: Polynomial[C], y: Polynomial[C]): Polynomial[C] = (x /~ gcd(x, y)) * y //FIXME?
```
tixxit: I think this is fine - we know that the remainder will be 0, so using `/~` is legit.

core/shared/src/main/scala/spire/math/package.scala
```diff
@@ -403,10 +403,126 @@ package object math {
...
+  // BigInt
+  /**
+   * This will return the largest integer that meets some criteria. Specifically,
+   * if we're looking for some integer `x` and `f(x')` is guaranteed to return
+   * `true` iff `x' <= x`, then this will return `x`.
+   *
+   * This can be used, for example, to find an integer `x` s.t. 
+   * `x * x < y < (x+1)*(x+1)`, by using `intSearch(x => x * x <= y)`.
+   */
+  private def intSearch(f: Int => Boolean): Int = {
```
tixxit: I only see this as an addition, without the corresponding existing code deleted. Does either this or the old version need to be removed?

core/shared/src/main/scala/spire/math/package.scala
```diff
@@ -403,10 +403,126 @@ package object math {
...
+   *
+   * See http://en.wikipedia.org/wiki/Shifting_nth_root_algorithm
+   *
+   * @param a A (positive if k % 2 == 0) `BigDecimal`.
+   * @param k A positive `Int` greater than 1.
+   * @param ctxt The `MathContext` to bound the precision of the result.
+   *
+   * returns A `BigDecimal` approximation to the `k`-th root of `a`.
+   */
+  def nroot(a: BigDecimal, k: Int, ctxt: MathContext): BigDecimal =
```
tixxit: Do we still use this? I can't remember if I wrote this or not. I'm wondering how the [implementation in `Algebraic`](https://github.com/non/spire/blob/master/core/shared/src/main/scala/spire/math/Algebraic.scala#L1098) compares. IIRC, this one was pretty slow, which is why we used to Newton's method for powers of 2.

tixxit: Nothing major here (and feel free to do nothing), but it may be that we can just delete this code entirely and use `Algebraic.nroot(a.bigDecimal, k, mc)` instead, for the time being. If you actually use `Algebraic(a).nroot(k).toBigDecimal(mc)` instead, you'll get exact rounding, but it'll be a bit slower and probably not worth it, given you can't expect much correctness with BigDecimal.

core/shared/src/main/scala/spire/std/bigDecimal.scala
```diff
@@ -77,7 +78,7 @@ trait BigDecimalIsNRoot extends NRoot[BigDecimal] {
   def nroot(a: BigDecimal, k: Int): BigDecimal = {
     if (a.mc.getPrecision <= 0)
       throw new ArithmeticException("Cannot find the nroot of a BigDecimal with unlimited precision.")
-    NRoot.nroot(a, k, a.mc)
+    spire.math.nroot(a, k, a.mc)
```
tixxit: Ah, cool, so we should definitely remove the existing code in `NRoot`. Or, I'd suggest we maybe use `Algebraic.nroot(a.bigDecimal, k, a.mc)` instead and delete the the shifting n-root algo entirely.

core/shared/src/main/scala/spire/std/int.scala
```diff
@@ -73,7 +74,7 @@ class IntIsBitString extends BitString[Int] with Serializable {
   def and(a: Int, b: Int): Int = a & b
   def or(a: Int, b: Int): Int = a | b
   def complement(a: Int): Int = ~a
-  override def xor(a: Int, b: Int): Int = a ^ b
+  //override def xor(a: Int, b: Int): Int = a ^ b
```
tixxit: Delete.

core/shared/src/main/scala/spire/std/int.scala
```diff
@@ -1,18 +1,18 @@
...
 import java.lang.Integer
 
-trait IntIsEuclideanRing extends EuclideanRing[Int] {
+trait IntIsEuclideanRing extends EuclideanRing[Int] with Gcd[Int] {
   override def minus(a:Int, b:Int): Int = a - b
   def negate(a:Int): Int = -a
   def one: Int = 1
   def plus(a:Int, b:Int): Int = a + b
-  override def pow(a:Int, b:Int): Int = spire.math.pow(a, b).toInt
+  //override def pow(a:Int, b:Int): Int = Math.pow(a, b).toInt //FIXME
```
tixxit: Delete.

macros/src/main/scala/spire/macros/Syntax.scala
```diff
@@ -3,14 +3,33 @@ package macros
...
 import spire.macros.compat.{termName, freshTermName, resetLocalAttrs, Context, setOrig}
 
-import scala.language.higherKinds
-
 object Ops extends machinist.Ops {
 
   def uesc(c: Char): String = "$u%04X".format(c.toInt)
 
   val operatorNames: Map[String, String] =
     machinist.DefaultOps.operatorNames ++ Map(
```
tixxit: Did the set of defaults change?

tests/src/test/scala/spire/SyntaxTest.scala
```diff
@@ -53,7 +53,8 @@ class SyntaxTest extends SpireTests with Checkers with BaseSyntaxTest {
   test("IsReal syntax")(check(forAll { (a: Double) => testIsRealSyntax(a) }))
   test("Semigroup syntax")(check(forAll { (a: String, b: String) => testSemigroupSyntax(a, b) }))
   test("Monoid syntax")(check(forAll { (a: String, b: String) => testMonoidSyntax(a, b) }))
-  test("Group syntax")(check(forAll { (a: Int, b: Int) => testMonoidSyntax(a, b)(Group.additive[Int]) }))
+  //test("Group syntax")(check(forAll { (a: Int, b: Int) => testMonoidSyntax(a, b)(Group.additive[Int]) }))
```
tixxit: Delete this line.

tests/src/test/scala/spire/SyntaxTest.scala
```diff
@@ -263,38 +265,36 @@ trait BaseSyntaxTest {
...
+      (-a == Ring[A].negate(a)) &&
+      ((a * b) == Ring[A].times(a, b)) &&
+      ((a /~ b) == EuclideanRing[A].quot(a, b)) &&
+      ((a % b) == EuclideanRing[A].mod(a, b)) &&
+      ((a /% b) == EuclideanRing[A].quotmod(a, b)) &&
+      ((a / b) == Field[A].div(a, b)) &&
+      ((a ** 2) == Ring[A].pow(a, 2)) &&
+      ((a pow 2) == Ring[A].pow(a, 2)) &&
+      // ((a gcd b) == Gcd[A].gcd(a, b)) &&
+      // ((a lcm b) == Gcd[A].lcm(a, b)) &&
```
tixxit: Why are these commented out?

tixxit:
I did a first pass. A few minor things, but overall looks really great. Thanks again!

core/shared/src/main/scala/spire/algebra/Gcd.scala
non: Good point. I'll remove the `implicit`.

core/shared/src/main/scala/spire/algebra/Gcd.scala
non: +1

core/shared/src/main/scala/spire/algebra/free/FreeGroup.scala
```diff
@@ -69,9 +69,9 @@ object FreeGroup { companion =>
   final def lift[A](a: A): FreeGroup[A] = new FreeGroup[A](Vector(Right(a)))
 
   implicit def FreeGroupGroup[A]: Group[FreeGroup[A]] = new Group[FreeGroup[A]] {
-    def id: FreeGroup[A] = companion.id
-    def op(a: FreeGroup[A], b: FreeGroup[A]): FreeGroup[A] = a |+| b
+    def empty: FreeGroup[A] = companion.id
+    def combine(a: FreeGroup[A], b: FreeGroup[A]): FreeGroup[A] = a |+| b
```
johnynek: you could have a nice `combineAll` here right (vector concatenation)?

core/shared/src/main/scala/spire/algebra/free/FreeMonoid.scala
```diff
@@ -11,7 +11,7 @@ final class FreeMonoid[A] private (val terms: List[A]) extends AnyVal { lhs =>
   def runSemigroup[B](f: A => B)(implicit B: Semigroup[B]): Option[B] =
     terms match {
       case head :: tail =>
-        Some(tail.foldLeft(f(head)) { (acc, a) => B.op(acc, f(a)) })
+        Some(tail.foldLeft(f(head)) { (acc, a) => B.combine(acc, f(a)) })
```
johnynek: don't you want: `B.combineAllOption(terms.iterator.map(f))`?

core/shared/src/main/scala/spire/algebra/free/FreeMonoid.scala
```diff
@@ -20,7 +20,7 @@ final class FreeMonoid[A] private (val terms: List[A]) extends AnyVal { lhs =>
    * Map each term to type `B` and sum them using `B`'s [[Monoid]].
    */
   def run[B](f: A => B)(implicit B: Monoid[B]): B =
-    terms.foldLeft(B.id) { (acc, a) => B.op(acc, f(a)) }
+    terms.foldLeft(B.empty) { (acc, a) => B.combine(acc, f(a)) }
```
johnynek: `B.combineAll(terms.iterator.map(f))`

core/shared/src/main/scala/spire/std/iterable.scala
```diff
@@ -9,8 +9,8 @@ import spire.algebra.Monoid
 @SerialVersionUID(0L)
 final class IterableMonoid[A, SA <: TraversableLike[A, SA]](implicit cbf: CanBuildFrom[SA, A, SA])
 extends Monoid[SA] with Serializable {
-  def id: SA = cbf().result()
-  def op(x: SA, y: SA): SA = x.++(y)(cbf)
+  def empty: SA = cbf().result()
+  def combine(x: SA, y: SA): SA = x.++(y)(cbf)
```
johnynek: can add `combineAllOption` on this.

core/shared/src/main/scala/spire/std/string.scala
```diff
@@ -5,8 +5,8 @@ import spire.algebra.{MetricSpace, Monoid, Order}
 
 @SerialVersionUID(0L)
 class StringMonoid extends Monoid[String] with Serializable {
-  def id: String = ""
-  def op(x: String, y: String): String = x + y
+  def empty: String = ""
+  def combine(x: String, y: String): String = x + y
```
johnynek: want `combineAll` on this for sure, otherwise N^2

core/shared/src/main/scala/spire/algebra/Gcd.scala
denisrosset: @tixxit If `EuclideanRing` and `Gcd` are defined properly, there will be no edge cases when dealing with polynomials.

I will have a look later today. The definition of `EuclideanRing` in `algebra` is fine, I will have a look at the proposed `Gcd` typeclass --- and all the default instances defined.

I will also make sure the documentation is clear, because:
- the mathematical definitions on Wikipedia (our default encyclopedia) are extremely rich (commutative rings ⊃ integral domains ⊃ integrally closed domains ⊃ GCD domains ⊃ unique factorization domains ⊃ principal ideal domains ⊃ Euclidean domains ⊃ fields ⊃ finite fields), not joking!
- the function `fmod(lhs, rhs)` does not correspond to a definition in the chain above (I interpret `fmod` as: return `lhs - k * rhs` where `k` is the maximal integer such that the result is positive). Still, we call often it "modulo", and do not see the difference with the modulo operation that appears in Euclid's algorithm.

tixxit: Here is the most recent fix somebody had for `Polynomial`'s gcd: #558

The default algorithm here works for integers, but not really rationals, where we want to find a common denominator first. In these cases, we override the definition of gcd to account for this.

I think, more generally, if we have this type class and it is open for people to implement/override themselves, perhaps for correctness or maybe just for speed, then having this default be implicit is dangerous, since it is super easy to silently replace a user's instance with this default.

tixxit: Also - really appreciate you taking a look at the implementations. In particular, the implementations for `Rational`, `Float`, and `Double` are somewhat involved. If the `Polynomial` one can be done w/ just euclid (perhaps w/ some adjustments to `euclid`), that'd be great. I'm not super clear on why the special case is necessary :\

tests/src/test/scala/spire/SyntaxTest.scala
non: Because `lcm` and `gcd` no longer come with `Field` for free.

macros/src/main/scala/spire/macros/Syntax.scala
non: Yeah I think we started enabling `-language:higherKinds` in `build.sbt`.

core/shared/src/main/scala/spire/std/bigDecimal.scala
non: I deferred (for now) on the question of using `Algebraic`.

non:
The PR hasn't made it obvious but I took all of @johnynek's suggestions.

core/shared/src/main/scala/spire/std/bigDecimal.scala
tixxit: Great - I'll add an issue for now and take a look shortly!

macros/src/main/scala/spire/macros/Syntax.scala
tixxit: Ah - I was more referring to adding all the operator mappings below :)

build.sbt
```diff
@@ -212,17 +212,14 @@ lazy val buildSettings = Seq(
...
     "-Ywarn-numeric-widen",
     "-Ywarn-value-discard"
   )),
-  resolvers ++= Seq(
-    "bintray/non" at "http://dl.bintray.com/non/maven",
-    Resolver.sonatypeRepo("snapshots")
-  ),
-  libraryDependencies += "org.typelevel" %%% "machinist" % "0.4.1"
+  libraryDependencies += "org.typelevel" %%% "machinist" % "0.5.0",
+  libraryDependencies += "org.typelevel" %%% "algebra" % "0.5.2-SNAPSHOT"
```
tixxit: I see this is a SNAPSHOT - are there any pre-requisites on algebra before merging this (aside from a release)?

core/shared/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -490,14 +485,17 @@ with EuclideanRing[Polynomial[C]] with VectorSpace[Polynomial[C], C] {
...
 
   final def gcd(x: Polynomial[C], y: Polynomial[C]): Polynomial[C] = {
-    val result = euclid(x, y)(Polynomial.eq)
+    val result = Gcd.euclid(x, y)(Polynomial.eq, this)
     if (result.degree > 0) {
       result
     } else {
       // return the gcd of all coefficients when there is no higher degree divisor
-      Polynomial.constant(spire.math.gcd(x.coeffsArray ++ y.coeffsArray))
+      val g = Gcd.fromEuclideanRing[C]
```
tixxit: Could we split out `Gcd[Polynomial[C]]` into its own, separate implicit instead? That way we can ask for a `Gcd[C]` instead.

```
class PolynomialGcd[C](implicit ring: EuclideanRing[Polynomial[C]], coeffGcd: Gcd[C]) extends Gcd[Polynomial[C]] {
  ....
}
```

build.sbt
non: No, I think we just need another algebra release. I wanted to wait to see if we needed other algebra changes though.

core/shared/src/main/scala/spire/math/Polynomial.scala
non: Sure.

denisrosset:
Regarding `EuclideanRing` and `Gcd`: we need to separate the division-with-remainder, and the `fmod` operations. I have a summary here, please tell me if you need any clarifications:

https://github.com/denisrosset/algebra/blob/master/EuclidianRing.md

johnynek:
Woo! Algebird merged today too: https://github.com/twitter/algebird/pull/523

Now we should coordinate the next release so we can push the one true common algebra core!

denisrosset:
@johnynek fantastic!

PR #577, opened 2016-09-13
Overlap operation for Intervals

vpavkin:
Following [gitter discussion](https://gitter.im/non/spire?at=57d6ff2fa0e8dbb04f308eab), I'm proposing a rich overlap operation for Intervals.

For any two Intervals `x` and `y`, overlap operation can return one of the following:

- `StrictlyLess(x , y)`, meaning that both are nonEmpty, `x ∩ y = Ø` and `x.upperBound < y.lowerBound`. Example: `(0, 1)` and `[3, 4]`.
- `LessAndOverlap(x , y)`. Partial overlap: both are nonEmpty, `y ∋ x.upperBound` && `y ∌ x.lowerBound`. Example: `[0, 4)` and `[3, 5)`
- `Subset(x, y)`, meaning `x` is a subset of `y` (including `x = Ø`).
- Any of the previous with `x` and `y` switched.
- `Equals(x, y)`, meaning `x` is equal to `y`

Also added `Eq` instances for `Bound` and `Overlap`.

P.S. Not sure, if actual overlap implementation is the best one performance-wise, advice is welcome :)

### DIFF
--- a/core/shared/src/main/scala/spire/math/Interval.scala
+++ b/core/shared/src/main/scala/spire/math/Interval.scala
@@ -7,6 +7,7 @@ import spire.algebra._
 import spire.math.poly.Term
 import spire.math.interval._
 import spire.math.interval.Bound._
+import spire.math.interval.Overlap._
 import spire.syntax.field._
 import spire.syntax.nroot._
 import spire.syntax.order._
@@ -771,28 +772,30 @@ sealed abstract class Interval[A](implicit order: Order[A]) extends Serializable
   /**
     * Result of overlapping this interval with another one.
     * Can be one of the following:
-    * - [[spire.math.interval.Equals]] if intervals are equal
-    * - [[spire.math.interval.StrictlyLess]] if intervals are notEmpty don't intersect
-    * - [[spire.math.interval.LessAndOverlaps]] if intervals intersect and neither is a subset of another
-    * - [[spire.math.interval.Subset]] if one interval (possibly empty) is a subset of another
+    * - [[Equal]] if intervals are equal
+    * - [[Disjoint]] if intervals are notEmpty don't intersect
+    * - [[PartialOverlap]] if intervals intersect and neither is a subset of another
+    * - [[Subset]] if one interval (possibly empty) is a subset of another
     *
-    * Note that lhs and rhs can be switched places in any result, except [[spire.math.interval.Equals]].
+    * Except for [[Equal]], both original intervals are bound to respective result fields,
+    * allowing to determine exact overlap type.
+    *
     * For example (pseudo-code):
     * {
-    * val lhs = [5, 6]
-    * val rhs = (0, 1)
+    * val a = [5, 6]
+    * val b = (0, 1)
     *
-    * // this returns StrictlyLess(rhs, lhs). Note that lhs and rhs switched places in compare to the overlap call
-    * lhs.overlap(rhs)
+    * // this returns Disjoint(b, a). Note a and b placement here, it means that b is strictly less then a.
+    * a.overlap(b)
     * }
     */
   def overlap(rhs: Interval[A]): Overlap[A] = {
 
     def lessAndOverlaps(intersectionLowerBound: Bound[A]): Overlap[A] =
-      if (lhs.lowerBound === intersectionLowerBound) LessAndOverlaps(lhs, rhs) else LessAndOverlaps(rhs, lhs)
+      if (lhs.lowerBound === intersectionLowerBound) PartialOverlap(lhs, rhs) else PartialOverlap(rhs, lhs)
 
     if (lhs === rhs) {
-      Equals(lhs, rhs)
+      Equal()
     } else {
       (lhs, rhs) match {
         case (sub, sup) if sup.isSupersetOf(sub) => Subset(sub, sup)
@@ -802,8 +805,8 @@ sealed abstract class Interval[A](implicit order: Order[A]) extends Serializable
           case i: Bounded[A] => lessAndOverlaps(i.lowerBound)
           case i: Point[A] => lessAndOverlaps(i.lowerBound)
           case Empty() =>
-            if (Interval.fromBounds(lhs.lowerBound, rhs.upperBound).isEmpty) StrictlyLess(rhs, lhs)
-            else StrictlyLess(lhs, rhs)
+            if (Interval.fromBounds(lhs.lowerBound, rhs.upperBound).isEmpty) Disjoint(rhs, lhs)
+            else Disjoint(lhs, rhs)
           case _ => throw new Exception("impossible")
         }
       }
--- a/core/shared/src/main/scala/spire/math/Interval.scala
+++ b/core/shared/src/main/scala/spire/math/Interval.scala
@@ -7,7 +7,6 @@ import spire.algebra._
 import spire.math.poly.Term
 import spire.math.interval._
 import spire.math.interval.Bound._
-import spire.math.interval.Overlap._
 import spire.syntax.field._
 import spire.syntax.nroot._
 import spire.syntax.order._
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/spire/math/Interval.scala
```diff
@@ -767,6 +767,48 @@ sealed abstract class Interval[A](implicit order: Order[A]) extends Serializable
 
   def foldOver[B](init: B, step: A)(f: (B, A) => B)(implicit ev: AdditiveMonoid[A], nt: NumberTag[A]): B =
     iterator(step).foldLeft(init)(f)
+
+  /**
+    * Result of overlapping this interval with another one.
+    * Can be one of the following:
+    * - [[spire.math.interval.Equals]] if intervals are equal
+    * - [[spire.math.interval.StrictlyLess]] if intervals are notEmpty don't intersect
```
tixxit: It seems a bit awkward that, eg, `lhs.overlap(rhs)` would return `StrictlyLess`, if `rhs` is strictly less than `lhs`. I wonder if maybe a better name would be something like `Disjoint`, instead of `StrictlyLess`. The order of the parameters in `Disjoint` would still maintain the order (eg 1st arg < 2nd arg), but the name seems to match the intent of the result a bit better.

core/shared/src/main/scala/spire/math/Interval.scala
```diff
@@ -767,6 +767,48 @@ sealed abstract class Interval[A](implicit order: Order[A]) extends Serializable
...
+    * - [[spire.math.interval.LessAndOverlaps]] if intervals intersect and neither is a subset of another
```
tixxit: Similar here - wonder if we could rename to `Overlaps` or `PartialOverlap` or something.

core/shared/src/main/scala/spire/math/interval/Overlap.scala
```diff
@@ -0,0 +1,90 @@
...
+  def isLessAndOverlaps: Boolean
+  def isSubset: Boolean
+  def isEqual: Boolean
+}
+
+/**
+  * Intervals are nonEmpty and don't intersect
+  * [[lhs.upperBound]] is strictly less than [[rhs.lowerBound]].
+  */
+case class StrictlyLess[A: Order] private[spire](lhs: Interval[A], rhs: Interval[A]) extends Overlap[A] {
```
tixxit: Could we move these case classes inside the companion object of `Overlap`? That seems to be the prevailing style now, even though we've been inconsistent in Spire historically.

core/shared/src/main/scala/spire/math/Interval.scala
```diff
@@ -767,6 +767,48 @@ sealed abstract class Interval[A](implicit order: Order[A]) extends Serializable
...
+    * For example (pseudo-code):
+    * {
+    * val lhs = [5, 6]
+    * val rhs = (0, 1)
+    *
+    * // this returns StrictlyLess(rhs, lhs). Note that lhs and rhs switched places in compare to the overlap call
+    * lhs.overlap(rhs)
+    * }
+    */
+  def overlap(rhs: Interval[A]): Overlap[A] = {
```
tixxit: We can use this to define a `PartialOrder` on `Interval` (where <= is defined for `StrictlyLess` and `Equals`) - may be nice to add that instance here.

core/shared/src/main/scala/spire/math/interval/Overlap.scala
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+  * An ADT that represents overlapping result for any two intervals.
+  * All subclasses and method names are "left-biased":
+  * for example, [[StrictlyLess]] result means, that [[lhs]] is strictly less than [[rhs]].
+  */
+sealed abstract class Overlap[A: Order] extends Serializable {
+  def lhs: Interval[A]
+  def rhs: Interval[A]
+  def isStrictlyLess: Boolean
```
tixxit: I think these could all be defined up-front here, with some pattern matching. I wonder if we also need all 4 - seems like the important ones `isDisjoint`, `isSubset`, and `isEqual`. If it is LessAndOverlaps or StrictlyLess, then both `isEqual` and `isSubset` will be false, and `isDisjoint` indicates the particular case.

core/shared/src/main/scala/spire/math/interval/Overlap.scala
```diff
@@ -0,0 +1,90 @@
...
+import spire.algebra.{Eq, Order}
+import spire.math.Interval
+
+/**
+  * An ADT that represents overlapping result for any two intervals.
+  * All subclasses and method names are "left-biased":
+  * for example, [[StrictlyLess]] result means, that [[lhs]] is strictly less than [[rhs]].
+  */
+sealed abstract class Overlap[A: Order] extends Serializable {
+  def lhs: Interval[A]
```
tixxit: I think `lhs` and `rhs` are bit oddly named, given that we may re-order the arguments from the original `overlap` call. Personal preference is that we actually get rid of `lhs` and `rhs` completely and expect folks to pattern match. If someone just wants to "order" them, then if we define the `PartialOrder[Interval[A]]` using the `overlap` method, then we can just order them with that instead.

tixxit:
Thanks for the work here! This definitely seems very useful to have. I suspect @non will also want to review, since he knows much more about `Interval` than me.

core/shared/src/main/scala/spire/math/Interval.scala
tixxit: The alternative is that we add 2 more cases to deal with `GreaterAndOverlap` and `StrictlyGreater` and we respect the original order of the args. In this case, we can directly use the result to determine the ordering of the intervals.

non: +1 on a name like `Disjoint`.

vpavkin: @tixxit, thanks!
Originally I had `StrictlyGreater` and other counterparts, but I more like your other suggestion about non-biased names with more descriptive interval names instead of `lhs`/`rhs`

core/shared/src/main/scala/spire/math/Interval.scala
vpavkin: Sure, should I put it in `spire.optional` similarly to `intervalGeometricPartialOrder` ?

vpavkin: actually, seems like `intervalGeometricPartialOrder` does pretty much the same thing...

tixxit: Ah - it exists already! Nevermind then :)

vpavkin:
@tixxit looks like I handled all the issues you mentioned. I looks much better now, thanks for good advice! 👍

core/shared/src/main/scala/spire/math/interval/Overlap.scala
```diff
@@ -0,0 +1,65 @@
+package spire.math.interval
+
+import spire.algebra.{Eq, Order}
+import spire.math.Interval
+import spire.math.interval.Overlap.{Disjoint, Equal, Subset}
+
+/**
+  * An ADT that represents overlapping result for any two intervals.
+  */
+sealed abstract class Overlap[A: Order] extends Serializable {
```
tixxit: We should avoid trapping/requiring an `Order` here. You'll need to add an `(implicit ord: Order[A])` parameter to `def join`, but otherwise I think everything should just work.

core/shared/src/main/scala/spire/math/interval/Overlap.scala
```diff
@@ -0,0 +1,65 @@
...
+  def isDisjoint: Boolean = this.isInstanceOf[Disjoint[_]]
+  def isSubset: Boolean = this match {
+    case Subset(_, _) | Equal() => true
+    case _ => false
+  }
+  def isEqual: Boolean = this.isInstanceOf[Equal[_]]
+}
+
+object Overlap {
```
tixxit: I should've thought of this in the last review (sorry!), but it feels like the `Interval#overlap` method implemented should live here instead as `apply[A: Order](a: Interval[A], b: Interval[A]): Overlap[A] = { ... }`. We'd still keep the `overlap` method on `Interval`, but it would just be implemented using `Overlap[A](this, that)` (eg `def overlap(that): Overlap[A] = Overlap(this, that)`).

Do you think that makes sense? Mainly, if we have this nice `Overlap` ADT, I feel like we should have some sort of public constructor on it and just a convenience method on `Interval` instead.

core/shared/src/main/scala/spire/math/interval/Overlap.scala
tixxit: Also, super minor, but we should also extend `Product` here (eg `extends Product with Serializable`) as per http://underscore.io/blog/posts/2015/06/04/more-on-sealed.html

Not a big deal - I think we should probably do this throughout Spire, so if it isn't done now, it'll probably happen in a later PR anyways.

tixxit:
@vpavkin I had 2 more suggestions - sorry for not brining them up earlier! Overall this looks great though - thanks for everything!

vpavkin:
no worries, I agree on both 👍

tixxit:
👍 - but we should wait on @non to review as well.

core/shared/src/main/scala/spire/math/interval/Overlap.scala
```diff
@@ -0,0 +1,89 @@
...
+
+    def lessAndOverlaps(intersectionLowerBound: Bound[A]): Overlap[A] =
+      if (lhs.lowerBound === intersectionLowerBound) PartialOverlap(lhs, rhs) else PartialOverlap(rhs, lhs)
+
+    if (lhs === rhs) {
+      Equal()
+    } else {
+      (lhs, rhs) match {
+        case (sub, sup) if sup.isSupersetOf(sub) => Subset(sub, sup)
+        case (sup, sub) if sup.isSupersetOf(sub) => Subset(sub, sup)
```
non: I think this outer pattern match could be better-written as:

```scala
if (rhs.isSupersetOf(lhs)) Subset(lhs, rhs)
else if (lhs.isSupersetOf(rhs)) Subset(rhs, lhs)
else {
  ...
}
```

It seems a bit nit-picky, but when I was reviewing this code I missed the fact that `sub` and `sup` were switched in the two cases, and thought the code was the same. (Those names are also very similar so I think it's easy to get confused.) It doesn't seem like the pattern-match is saving us any code, so maybe we can just use `lhs` and `rhs` with if/else branches?

non:
I had one comment. After that, 👍 from me as well.

Thanks for adding this! I think this will be a really useful feature for everyone working with intervals.

vpavkin:
@non thanks, fixed! totally reasonable comment :)

non:
Looks great! Thanks again. 👍

PR #578, opened 2016-09-21
Root isolation failure test case.

denisrosset:
Test case reporting the following bug:
```scala
scala> import spire.math._; import spire.implicits._
import spire.math._
import spire.implicits._

scala> Algebraic.roots(Polynomial("4x^3 + 2x^2 - 3x - 1"))
res0: Vector[spire.math.Algebraic] = Vector(Algebraic(-1), Algebraic(-1), Algebraic(~0.8090169943749474))
```

where actually the roots should be -1, (1 - sqrt(5))/4 and (1 + sqrt(5))/4

Note: I'm planning to investigate this later. The problem comes probably from the confusion between open and closed intervals, when isolating the root.

### DIFF
--- a/core/shared/src/main/scala/spire/math/poly/BigDecimalRootRefinement.scala
+++ b/core/shared/src/main/scala/spire/math/poly/BigDecimalRootRefinement.scala
@@ -136,10 +136,10 @@ object BigDecimalRootRefinement {
         .round(mc)
 
     def floor(x: Rational): JBigDecimal =
-      x.toBigDecimal(new MathContext(mc.getPrecision, RoundingMode.CEILING)).bigDecimal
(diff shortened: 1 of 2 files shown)


tixxit:
I'll take a look too - there should only be open intervals for the root refinement, so if we ever consider the lower or upper bound a candidate for the root, then it is a bug/mistake.

tixxit:
Ooooh - I think I found the problem. I'll make a PR against your PR if this fixes it...

tixxit:
Here is the fix: https://github.com/denisrosset/spire/pull/2

denisrosset:
Thanks a lot! I'd still like to write additional randomized tests for root isolation before merging.

non:
Sounds good. Thanks @denisrosset and @tixxit.

denisrosset:
Superceded by #579 
Thanks @tixxit

PR #579, opened 2016-09-25
Bug/various root isolation and refinement bugs

tixxit:
This fixes a few bugs:

1) A bug in the detection of roots in between an exact rational bound
and a very close approximation. Usually the root would fall in between 2
approximations, and even then, in this specific case, the root detection
usually still reported the correct result.

2) Bug in the transformation of a root upper bound to an isolated root
bound. Originally, it was using the inverse Mobius transform, but
should've been using the Mobius transform as-is.

3) A bug in the computation the upper bound of real roots of a
polynomial. This was an off-by-1 bug that affected cases where
roots fell very near the actual upper bound - within 1 bit - and
the largest coefficient was on a term of degree <= 2.

Most importantly, this re-enables the rational roots property test for
Algebraic, which was the scalacheck test that found all 3 of these bugs.

@denisrosset This also includes your broken test + my fix for that.

### DIFF
--- a/core/shared/src/main/scala/spire/math/poly/BigDecimalRootRefinement.scala
+++ b/core/shared/src/main/scala/spire/math/poly/BigDecimalRootRefinement.scala
@@ -136,10 +136,10 @@ object BigDecimalRootRefinement {
         .round(mc)
 
     def floor(x: Rational): JBigDecimal =
-      x.toBigDecimal(new MathContext(mc.getPrecision, RoundingMode.CEILING)).bigDecimal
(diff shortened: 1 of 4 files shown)


tixxit:
Also, as a sad note, I accidently obliterated (`git checkout` :() the extracted test cases for 2 of the bugs. If they existed, then the rational roots property test would find them fairly predictably in a few runs anyways, so I'm fairly sure we won't see regressions, but still a bit sad.

non:
Looks good to me! 👍

PR #581, opened 2016-09-27
Improve performance of Polynomial#shift.

tixxit:
This improves the performance of Polynomial's `shift` method, which is the major bottleneck in the VAS root isolation algorithm. It uses the trick described here: http://math.stackexchange.com/questions/694565/polynomial-shift

The tl;dr is that when I ran the "find all rational roots of rational polynomial" test w/ minSuccessful(1000), it took over a minute to run before. After this update it takes ~15s, which is a pretty substantial improvement.

### DIFF
--- a/core/shared/src/main/scala/spire/math/Polynomial.scala
+++ b/core/shared/src/main/scala/spire/math/Polynomial.scala
@@ -3,9 +3,6 @@ package math
 
 import scala.collection.mutable.ArrayBuilder
 
-
-import java.math.{ BigDecimal => JBigDecimal, RoundingMode, MathContext }
-
 import spire.algebra._
 import spire.math.poly._
 import spire.std.array._
@@ -47,8 +44,6 @@ object Polynomial extends PolynomialInstances {
 
   def apply[@sp(Double) C: Semiring: Eq: ClassTag](c: C, e: Int): PolySparse[C] =
     PolySparse.safe(Array(e), Array(c))
-
-  import scala.util.{Try, Success, Failure}
 
   def apply(s: String): Polynomial[Rational] = parse(s)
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -285,6 +280,59 @@ trait Polynomial[@sp(Double) C] { lhs =>
...
+   * required. This will require work though, so another time.
+   */
+  def shift(h: C)(implicit ring: Ring[C], eq: Eq[C]): Polynomial[C] = {
+    // The trick here came from this answer:
+    //   http://math.stackexchange.com/questions/694565/polynomial-shift
+    // This is a heavily optimized version of the same idea. This is fairly
+    // critical method to be fast, since it is the most expensive part of the
+    // VAS root isolation algorithm.
+
+    def fromSafeLong(x: SafeLong): C =
```
tixxit: How would people feel about possibly adding `fromBigInt` to `Ring` in algebra? We can use something like below to write a default implementation, but override it for, eg, BigInt, SafeLong, Rational, Algebraic, BigDecimal, etc.

core/shared/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -285,6 +280,59 @@ trait Polynomial[@sp(Double) C] { lhs =>
...
+      }
+
+    val coeffs: Array[C] = this.coeffsArray.clone()
+    this.foreachNonZero { (deg, c) =>
+      var i = 1
+      var d = deg - 1
+      var m = SafeLong(1L)
+      var k = c
+      while (d >= 0) {
+        m = (m * (d + 1)) /~ i
```
tixxit: Note that we do division, but only on SafeLongs. This is not just for performance, but also required for us to only ask for a `Ring`, rather than a `EuclideanRing`. We always know that `m * (d + 1)` is divisible by `i`, so this is exact.

I should probably add this comment to the code...

core/shared/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -326,22 +374,8 @@ trait Polynomial[@sp(Double) C] { lhs =>
...
-   * `poly.compose(x + h)`.
-   */
-  def shift(h: C)(implicit ring: Rig[C], eq: Eq[C]): Polynomial[C] =
-    compose(Polynomial.x[C] + Polynomial.constant(h))
-
-  /**
-   * Translates this polynomial by `h`. Equivalent to calling
-   * `poly.compose(x - h)`.
-   */
-  def translate(h: C)(implicit ring: Ring[C], eq: Eq[C]): Polynomial[C] =
```
tixxit: I removed `translate`, since it was basically another name for `shift`. Happy for other opinions. It wasn't being used anywhere though.

core/shared/src/main/scala/spire/math/Polynomial.scala
non: Sounds good to me.

tixxit: Cool - I'll leave this in here, but make a PR against algebra tonight. Once we move over to algebra, I'll just re-write this method to use it instead.

core/shared/src/main/scala/spire/math/Polynomial.scala
non: Seems fine to me.

core/shared/src/main/scala/spire/math/Polynomial.scala
non: Sounds good.

non:
Modulo the additional comments (and algebra PR) this is great! 👍

tixxit:
I added a few comments and also made another minor-ish performance improvement for `mapTerms`. Mind taking a look again, @non ? If we merge this now, then I'll make sure we update to `fromBigInt` in algebra once that gets merged: https://github.com/typelevel/algebra/pull/173

non:
This looks good. 👍 

The test failure doesn't appear to be related to this. I'm leaving this open in case you want to tack a fix on, but I'm also OK with merging this first.

tixxit:
Alright - fixed the broken test and things are green! Merging!

PR #582, opened 2016-09-30
Find a tighter lower bound in root isolation.

tixxit:
This is another perf improvement. This actually gets a pretty impressive improvement in the tests - well above 2x. Mainly, taking some time to find a tighter lower bound saves us quite a lot of calls to `Polynomial#shift`.


denisrosset:
LGTM.

Thanks a lot for these improvements. Note: I'm writing an add-on library for Spire to implement cyclotomics (github.com/denisrosset/spire-cyclo); I can restrict my code to real cyclotomics and have a total order. The tests on that total order are going to exercise the root-finding code intensively.

PR #583, opened 2016-09-30
compute negative nroots for Real numbers

pedropramos:
fixes https://github.com/non/spire/issues/496


tixxit:
Looks great! :+1:

non:
👍 Thanks again @pedropramos!

tixxit:
Weird error in travis - it timed out after not receiving output for 10 min. I restarted the tests, so we'll see if it was transient.

tixxit:
@pedropramos The tests "failed" again (> 10min w/ no output). When you run the tests locally, do they finish?

pedropramos:
@tixxit 
No, indeed my tests don't always finish. It seems that nroot goes into an infinite recursion when the k argument is -1. Unfortanely ScalaCheck did not generate a k=-1 for my first local test run.

The source of the bug is Rational.limitToInt (called by Real.fpow), which does not terminate when called with the value Rational(-1).

I have now rewritten nroot so that it doesn't call fpow directly. This should get the tests running successfully. I'll raise a separate issue for Rational.limitToInt.

PR #586, opened 2016-10-06
Fix slow convergence behaviour in Rational#limitTo.

### LINKED ISSUE #585: Rational.limitToInt doesn't terminate for 1 and -1
The following inputs lead to infinite recursion
```
Rational(-1).limitToInt
```
```
Rational(1).limitToInt
```

tixxit:
Fixes #585

This fixes a class of bugs in limitTo which lead to horrible performance in `Rational#limitTo`. Before, `Rational(1).limitToInt` would spin for a loooooong time before finally returning. Now it completes almost instantly.

### DIFF
--- a/core/shared/src/main/scala/spire/math/Algebraic.scala
+++ b/core/shared/src/main/scala/spire/math/Algebraic.scala
@@ -2,15 +2,13 @@ package spire
 package math
 
 import java.lang.Long.numberOfLeadingZeros
-import java.lang.Double.{ isInfinite, isNaN }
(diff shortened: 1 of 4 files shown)


denisrosset:
Wouldn't we prefer to use `Spire.util.Opt` in internal loops/methods, as they avoid allocating memory for `Opt[_ <: NotNull]`?

Apart from that, LGTM.

tixxit:
Ah, good point. I'll fix!

denisrosset:
Feel free to merge afterwards!

PR #587, opened 2016-10-08
use special Scala 2.11 sources for 2.12 too

SethTisue:
I don't know if copying these sources is the ideal way to resolve this
-- y'all can decide for yourselves. but I need some change along these
lines in order to resume tracking Spire's master branch in the Scala
2.12 community build.

note that CrossVersion.partialVersion will do the right thing even
for Scala milestones and RCs, weird dbuild-generated community
build Scala version numbers, etc.


dwijnand:
Here's one way to dedup if anyone is interested: https://github.com/milessabin/shapeless/blob/6cbe70e43e38a9c6f340aa0f2dc2065e6c8685a5/build.sbt#L200-L210

PR #588, opened 2016-10-10
Cleaned up imports.

denisrosset:
I split manually the patch of the #576 PR. This is the part that remove most of the warnings (import, interpolation, etc...). I however did not modify the flags in `build.sbt`.

The rest of the patch is here: https://gist.github.com/denisrosset/45774ba1cb9221a75f9c6be85f80fd4a

I'm working on importing the group terminology rename now.

### DIFF
--- a/benchmark/src/main/scala/spire/benchmark/AddBenchmarks.scala
+++ b/benchmark/src/main/scala/spire/benchmark/AddBenchmarks.scala
@@ -1,19 +1,9 @@
 package spire
 package benchmark
 
-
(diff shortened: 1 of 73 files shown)


denisrosset:
OK.. seems there is a randomized root isolation error.

denisrosset:
OK.. seems there is a randomized root isolation error.

```
[info] - find all rational roots of rational polynomial *** FAILED *** (506 milliseconds)
```

non:
LGTM 👍

PR #590, opened 2016-10-12
[READY] Lawful Signed

denisrosset:
Added laws for Signed, inspired by linearly ordered groups, but generalized to monoids to cater for the `UInt`, ... instances.

Added commutativity requirement where relevant (in Rig, UXXX instances).


denisrosset:
Due to `Signed` extending `Order` now, `Jet` is no longer a `Signed` instance.

Should we add a compare method on `Jet` ?

core/shared/src/main/scala/spire/math/Numeric.scala
```diff
@@ -229,7 +229,18 @@ private[math] class RealIsNumeric extends Numeric[Real] with RealIsFractional wi
   override def toDouble(n: Real): Double = n.toDouble
 }
 
+/* TODO: what to do with complex, which is not ordered?
```
tixxit: This is probably a good question for @non  - but given that `Numeric` just isn't used much, I'm OK with losing `Numeric[Complex[A]]`.

The alternative is that we just stub out the ordering or do something janky. `Numeric` basically exists as a "best" effort type class and it should be somewhat accepted that not all operations may exist/obey useful laws... Just typing that makes me think `Numeric` should just kind of disappear...

tixxit:
This LGTM, but I do think we should resolve the `Numeric[Complex[..]]` issue before merging. Either implement in a janky way or just get rid of it. I'm OK with getting rid of it, but others may feel differently.

core/shared/src/main/scala/spire/math/Numeric.scala
denisrosset: I don't want to touch `Numeric` before we complete the lawful typeclasses. In particular, `TruncatedDivision` and `UniqueFactorizationDomain`. I'd also like to find a way to represent homomorphisms (i.e. correspondances between types that preserve a structure).

I have in mind a list of typeclasses that should be precised; Numeric is not alone, see below.

In this thread [Haskell, the bad parts](https://www.reddit.com/r/haskell/comments/4sihcv/haskell_the_bad_parts/), the following comments are striking: "The numeric typeclasses are a mess." "Sure, but is there any language that gets them right?"

Let's get them right, but move steadily. I'm fine with removing `Numeric[Complex]` because it inherits `Signed` and `Order` through `IsReal`, which does not make sense for complex numbers. I'm currently *for* a removal a `Numeric`, but not before we made our best shots at implementing lawful alternatives.

(List of typeclasses to review)

- `IsReal`: part `TruncatedDivision`, part "contains the integers as subring --  `isWhole`", part "subfield of the reals (`toDouble`, `toReal`),

- `IsIntegral`, `IsRational`, `IsAlgebraic` should be handled by instances of `Homomorphism[A, B, F[_]] <: Function1[A, B]` where `F` is the structure being preserved (ring or field),

- the `ConvertableXXX` stuff is basically homorphisms, or approximations of these,

- `spire.math.Integral`: part `UniqueFactorizationDomain` which applies to polynomials as well (`factor`, `isPrime`), part helper methods (!, choose) that should be available for all types isomorphic to the integers (BigInt, BigInteger, SafeLong),

- `Fractional` (and `Integral`) seem to be modelled after their Haskell's equivalents. Imho, the numeric hierarchy of Haskell is not the greatest part of the language*.

tixxit:
OK, would you mind if we just deleted `Numeric[Complex[A]]` for now, instead of commenting it out. I'll create an issue for discussion and we can merge this!

PR #600, opened 2016-11-05
update machinist 0.6.1

xuwei-k:
I think we should update latest machinist if support Scala 2.12 https://github.com/non/spire/issues/596
- https://oss.sonatype.org/content/repositories/releases/org/typelevel/machinist_2.12/
- https://github.com/typelevel/machinist/commit/86dd0e4ff619b7871c7c54670b03fc4b24f001da

also fix https://github.com/non/spire/issues/592

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -281,7 +281,7 @@ in this guide.
 #### Groups
 
 The most basic algebraic type classes Spire supports involve a single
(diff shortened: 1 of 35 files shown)


denisrosset:
Let's do an interim 2.12 release before the cats-kernel -> algebra -> spire dependency introduction. There are enough small points to solve before introducing bigger changes.

Please look at #606 .

PR #605, opened 2016-11-10
Upgrade dependencies to build using scala 2.12

andrewsmartin:
For #596

PR #606, opened 2016-11-22
Preliminary work on Scala 2.12 update.

denisrosset:
Default on 2.12.0 build.
Updated dependencies.
Reverted machinist syntax to pre-cats-kernel notation.
Updated build plugins and SBT.


denisrosset:
Remaining blocking points:

1) optimizer settings: the inliner should be enabled only for production builds. Has anybody a `project.sbt` example on how to do so?

2) `-Yinline-warnings` is not available on the latest scala-js. Is this option needed (thus it should be conditionally enabled), or can we remove it?

3) some macro patterns have changed in 2.12. For example, `scala.this.Predef.intWrapper($i).until($j)` becomes `scala.Predef.intWrapper($i).until($j)`. How to deal with these changes? Special 2.12 code? Small compatibility layer?

denisrosset:
Thanks @andrewsmartin and @xuwei-k for preliminary work on this in #600 and #605.

tixxit:
@denisrosset This is looking good so far! Seems like some changes in Scalacheck are causing compilation issues.

For the -Yinline-warnings, you could probably add it here: https://github.com/non/spire/blob/master/build.sbt#L233

Those settings apply to the JVM only, so shouldn't cause problems w/ scalajs.

What do you mean about the inline only production builds? Right now we use `-optimize` in JVM builds, which should also enable inlining. Do we want to do something different?

For the last issue, I guess we could just add it to our existing compat layer stuff we have w/ the `scala_2XX` source dirs. There should be one for 2.12 that exists already, from Seth's PR.

denisrosset:
Regarding 1), I quote the Scala release notes:

```We recommend enabling inlining only in production builds, as sbt’s incremental compilation does not track dependencies introduced by inlining.```

**Comment by tixxit** [comment]
@denisrosset Got it! Let's create an issue for 1), but get the rest of the PR merged. SG?

**Comment by denisrosset** [comment]
@tixxit: do you know the logic behind `spire.macros.compat` and `HasCompat`?

in https://github.com/non/spire/blob/master/macros/src/main/scala/spire/macros/Syntax.scala#L77

**Comment by tixxit** [comment]
@denisrosset I don't know the reason for that - maybe @non can help?

**Comment by denisrosset** [comment]
@tixxit let's see. I added the `scala.Predef` compatibility hack to `compat`, without knowing exactly the import logic. It works, that's all.

I'm pushing some changes to adapt for Scalacheck. Three tests are failing, I have to find the mistake.

**Comment by denisrosset** [comment]
Okay, after updating Travis to JDK8, the CI should pass.

Will try to remove the deprecation warnings in build.sbt, and we should be good to go.

**Comment by tixxit** [comment]
@denisrosset Does Spire still build on Java 7?

**Comment by denisrosset** [comment]
@tixxit, did we already drop Java 6?

Can we do conditional CI so that 2.11 is tested on JDK 6 or 7, and 2.12 on JDK8 ?

**Comment by tixxit** [comment]
Ah, that looks great! (latest commit)

**Comment by tixxit** [comment]
👍  (assuming tests pass)

**Comment by denisrosset** [comment]
```
java.lang.NoClassDefFoundError: scoverage/Invoker$
	at spire.math.UShortTest.<init>(UnsignedTest.scala:221)
```

Crap. 

Some also have the same error:

http://blog.ssanj.net/posts/2016-01-14-scoverage-breaks-with-java-lang-NoClassDefFoundError-scoverage-Invoker.html

and recommend:

> @alexflav23 recommends turning off coverage when publishing or packaging:
>    You need to disable coverage with coverageOff before publishing or packaging and everything will be just fine. I’ve ran into this issue myself and that was the quick fix.

denisrosset:
> Ah, that looks great! (latest commit)

Shamelessly stolen from `cats`.

tixxit:
Ugh - these keep timing out failing w/ BS errors. I think one of the prime tests also tends to hang on ScalaJS :\

PR #614, opened 2016-11-30
[READY] Unique factorization domains for integers

denisrosset:
This is preliminary work to introduce [Unique factorization domains](https://en.wikipedia.org/wiki/Unique_factorization_domain) for integers.

In the future, this could be used in other domains such as polynomial rings.

### DIFF
--- a/core/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
+++ b/core/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
@@ -5,10 +5,10 @@ import spire.math.{Integral, SafeLong}
 
 /** A unique factorization domain is a commutative ring in which each element can be written
   * as a product of prime elements and a unit.
-  * 
+  *
   * Unique factorization domains are GCD rings (or domains), but not necessarily Euclidean
   * domains.
-  * 
+  *
   * This trait is outside the commutative ring hierarchy, because the factorization algorithms
   * are costly. Another reason: in some cases, a deliberate choice should be made by the user,
   * for example to use probabilistic algorithms with a specified probability of failure.
--- a/core/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
+++ b/core/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
@@ -1,0 +1,49 @@
+package spire
+package algebra
+
+import spire.math.{Integral, SafeLong}
+
+/** A unique factorization domain is a commutative ring in which each element can be written
+  * as a product of prime elements and a unit.
+  * 
+  * Unique factorization domains are GCD rings (or domains), but not necessarily Euclidean
+  * domains.
+  * 
+  * This trait is outside the commutative ring hierarchy, because the factorization algorithms
+  * are costly. Another reason: in some cases, a deliberate choice should be made by the user,
+  * for example to use probabilistic algorithms with a specified probability of failure.
+  */
+trait UniqueFactorizationDomain[@sp(Byte, Short, Int, Long) A] extends Any {
+  /** Tests whether the given nonzero element is prime. */
+  def isPrime(a: A): Boolean
+
+  /** Returns the factors of the given nonzero element. */
+  def factor(a: A): UniqueFactorizationDomain.Decomposition[A]
+}
+
+object UniqueFactorizationDomain {
+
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
```diff
@@ -0,0 +1,47 @@
...
+    def elements: Map[A, Int]
+  }
+
+  def apply[A](implicit ev: UniqueFactorizationDomain[A]): UniqueFactorizationDomain[A] = ev
+
+  case class WrapDecomposition[A:CRing](safeLongFactors: spire.math.prime.Factors) extends Decomposition[A] {
+    def unit: A = safeLongFactors.sign match {
+      case Sign.Negative => CRing[A].negate(CRing[A].one)
+      case Sign.Positive => CRing[A].one
+      case _ => sys.error("Factorization of zero is undefined.")
```
tixxit: Could this throw an `ArithmeticException` instead?

denisrosset: Yes, thanks!

core/shared/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
```diff
@@ -0,0 +1,47 @@
...
+object UniqueFactorizationDomain {
+
+  trait Decomposition[@sp(Byte, Short, Int, Long) A] {
+    def unit: A
+    def elements: Map[A, Int]
+  }
+
+  def apply[A](implicit ev: UniqueFactorizationDomain[A]): UniqueFactorizationDomain[A] = ev
+
+  case class WrapDecomposition[A:CRing](safeLongFactors: spire.math.prime.Factors) extends Decomposition[A] {
```
tixxit: Instead of a `WrapDecomposition`, maybe we just have a `map[B](f: A => B): Decomposition[B]` method in `Decomposition`? Then we can just define the `Integral` case as `def factor(a: A): Decomposition[A] = SafeLong(A.toBigInt(a)).factor.map(CRing[A].fromBigInt(_))`.

denisrosset: I actually need this map to be a homomorphism (for example, it would not work from `SafeLong` to `Rational`, as a field is a trivial factorization domain, all elements are units). I started work on homomorphisms, see #627.

When should we merge this?

denisrosset:
Ok... seems the factorization tests can be dramatically slow, depending on the seed. I will see what can be done.

tixxit:
@denisrosset Have you figured this out yet? Back when root finding was slower, I ended up just requiring fewer successful checks (like 10?) and also had my generators use small degree polynomials. I know it's definitely harder with the law tests, since we don't have as much control over the generator used, but maybe we could muck about w/ the size a bit...

denisrosset:
Mmmh.. not yet. I will probably define a "Factorizable" wrapper class that maps a "modulo maxSize" on the Arbitrary instance.

denisrosset:
I thought of the best way to handle this:

1. Replace the current `BigInt` generator from Scalacheck by a sized generator on the number of bits. Anyway, the Scalacheck generator seems to multiply Long-sized integers together, instead of generating 64-bit sized chunks.

2. Add a complexity predicate that estimates the test instance running time, and discards evaluations that would be too expensive.

3. Or, instead of discarding, shrinks appropriately the arguments.

I see how to incorporate 2. and 3. in Spire; but without support in Scalacheck/Scalatest, the reporting will be affected (either by failing tests due to insufficient completions, or by reporting random arguments before shrinking).

With appropriate support, the complexity budget could be estimated over all random instances and not just on a single run (so one of the tests can be expensive, but then the others have to be quite trivial). That would enable better coverage.

core/shared/src/main/scala/spire/algebra/UniqueFactorizationDomain.scala
denisrosset: I'll see what I can do reusing `Integral` before working on homomorphisms (which can be opening a can of worms).

denisrosset:
This has been rebased on the modularization work.

To fix later: tests for SafeLong/BigInt are too slow, so they are commented out for now.

To discuss with @non

non:
Looks fine to me. 👍

PR #617, opened 2017-01-02
Unsigned integer types form a commutative Rig.

denisrosset:
For some reason, this was missing.

The test coverage should be more complete for those types (modulo the redundancy in UnsignedTest)


tixxit:
👍

PR #618, opened 2017-01-02
DivisionRing and Quaternion made lawful.

denisrosset:
Added a DivisionRing typeclass.

Converted Quaternion to DivisionRing; split the Quaternion instances into one requiring only a signed Field scalar, and another richer that include roots but requires a NRoot/Trig scalar.

Added tests using the Quaternion[Rational] type.

Note: I changed some `IsReal` implicits into `Order` / `Signed` to be more relaxed in the requirements. This change will be complete in a future PR.

### DIFF
--- a/core/shared/src/main/scala/spire/algebra/DivisionRing.scala
+++ b/core/shared/src/main/scala/spire/algebra/DivisionRing.scala
@@ -1,0 +1,50 @@
+package spire
+package algebra
+
+import java.lang.Double.{ isInfinite, isNaN, doubleToLongBits }
+import java.lang.Long.{ numberOfTrailingZeros }
+
+trait DivisionRing[@sp(Byte, Short, Int, Long, Float, Double) A] extends Any with Ring[A] with MultiplicativeGroup[A] { self =>
+
+  def fromDouble(a: Double): A = DivisionRing.fromDouble[A](a)(self, self)
+}
+
+
+object DivisionRing {
+
+  @inline final def apply[A](implicit f: DivisionRing[A]): DivisionRing[A] = f
+
+  // TODO: use instead algebra Ring.fromDouble, if PR algebra/#190 is accepted
+  /**
+    * This is implemented in terms of basic ops. However, this is
+    * probably significantly less efficient than can be done with a specific
+    * type. So, it is recommended to specialize this general method.
+    *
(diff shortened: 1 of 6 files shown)


denisrosset:
Note: the codecov/patch failure is related to the fromDouble code duplication, which will come later from algebra `Ring.defaultFromDouble`.

core/shared/src/main/scala/spire/algebra/DivisionRing.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  def fromDouble(a: Double): A = DivisionRing.fromDouble[A](a)(self, self)
+}
+
+
+object DivisionRing {
+
+  @inline final def apply[A](implicit f: DivisionRing[A]): DivisionRing[A] = f
+
+  // TODO: use instead algebra Ring.fromDouble, if PR algebra/#190 is accepted
```
tixxit: This was merged! Can we delete this and use the one from here now? https://github.com/typelevel/algebra/pull/190

core/shared/src/main/scala/spire/math/Quaternion.scala
```diff
@@ -50,33 +44,47 @@ private[math] trait QuaternionAlgebra[A]
...
-    _gcd(a, b)
-  }
-   */
+
+  def timesl(a: A, q: Quaternion[A]): Quaternion[A] = q * a
+  def dot(x: Quaternion[A], y: Quaternion[A]): A = x.dot(y)
+
+}
+
+private[math] trait QuaternionOverRichField[A] extends QuaternionOverField[A]
```
tixxit: Nice split 👍

core/shared/src/main/scala/spire/algebra/DivisionRing.scala
denisrosset: typelevel/algebra#190 was merged but not published. Can Spire depend on snapshot versions of algebra?

denisrosset: Let's file an issue to keep track of this todo, and correct it later.

tixxit:
👍

denisrosset:
Thanks! I added an issue #622 to remind us the future code removal.

PR #619, opened 2017-01-02
Remove === and =!= methods (continuation of #527)

denisrosset:
Let's remove these `===` `=!=` misleading methods.


denisrosset:
I still kept the `===` and `=!=` shortcuts for the unsigned value classes, to avoid going through X method calls for a simple comparison.

To check: is the behavior of the default `==` (equals) correct when comparing a ULong and UInt for example?

denisrosset:
Closed as part of the umbrella issue.

PR #620, opened 2017-01-03
Added test for alias resolution.

denisrosset:
Checks that implicit resolution works across the Spire/Algebra/Cats boundaries.


tixxit:
👍

PR #627, opened 2017-02-01
[WIP] Homomorphisms.

denisrosset:


core/shared/src/main/scala/spire/algebra/Homomorphism.scala
```diff
@@ -0,0 +1,54 @@
+package spire.algebra
+
+import java.math.BigInteger
+
+import spire.math.{Complex, SafeLong}
+
+trait Homomorphism[-A, +B, +F[_]] extends Function1[A, B]
```
tixxit: I don't think this should extend `Function1`, since it means imported homomorphisms will also add implicit conversions which may be unexpected.

denisrosset: Good point!

denisrosset: @tixxit, homomorphisms do not extend `Function1` anymore. Instead, their implementation *could*, and they provide a `.toFunction` method.

denisrosset:
Next steps:

- [ ] add syntax for `Homomorphism`, something like `2.to[BigInt]`
- [ ] (when dealing with magnitude-limited types) integrate `NumberTag` to provide `.isValid[Int]`, `.isValid[Long]` and so on
- [ ] take a look at `Convertable`, `Integral`, `Fractional` and `Numeric`, see how they should evolve

 replace `ConvertableTo/From` by homomorphisms;

denisrosset:
Big question: should we call those "homomorphisms"?

If find "Convertable" much better to convey the meaning, as users are probably not super familiar with abstract algebra.

(OK for me to use nonstandard terminology, as long as we make use, behind the scene, of all the guarantees offered by the mathematical construction)

erikerlandson:
Big +1 for "Convertable" instead of "Homomorphism"    I saw this "homomophism" PR title several times and never once thought "he's working on numeric conversions."  😃

erikerlandson:
Or maybe define some kind of `Conversion` synonym.  Because now that I see `Homomorphism` I love it.

denisrosset:
I see a way out; `Conversion` is fine, with an additional type member `Hom` that specifies the homomorphism type --- this type member can be completely forgotten if you don't care about the algebraic laws.

@tixxit: I pushed some WIP that enables conversion from, and to, all integer types.

The syntax is (warning: bug in `hIsValid` when converting between two limited range types):

```scala
scala> import spire.math._; import spire.algebra._; import spire.implicits._
import spire.math._
import spire.algebra._
import spire.implicits._

scala> 200L.hTo[Int]
res1: Int = 200

scala> 200L.hTo[Short]
res2: Short = 200

scala> BigInt(1000).hTo[SafeLong]
res3: spire.math.SafeLong = 1000

scala> 100.hTo[java.math.BigInteger]
res4: java.math.BigInteger = 100

scala> 100.hIsValid[Byte] // hIsValid leverages `NumberTag`
res5: Boolean = true

scala> 200.hIsValid[Byte]
res6: Boolean = false
```

erikerlandson:
That seems intuitive.  A conversion makes use of a homomorphism to do its job.

denisrosset:
Closing for now

Issue #375, opened 2015-02-10
Have a type class with vector cross product for 3 dimensional space


pleira:
Spire could have a type class implementing the [vector cross product](http://en.wikipedia.org/wiki/Cross_product), which I would call something like EuclideanVectorSpace or EuclideanCoordinateSpace. This is an operation that is only suitable for 3 dimensional vector spaces and it is often used in physics. 

I have done a [preliminary implementation](https://github.com/pleira/spire/compare/EuclideanCoordinateSpace), offering a method angle as well. As the current CoordinateSpace [might be replaced](https://github.com/tixxit/spire/tree/topic/vectorspace-redux), it would be interesting to bring the cross product operation along with the replacement, or in a subproject from Spire.

denisrosset:
I had a look at the cross-product implementation. It requires the addition of a typeclass + syntax just for the cross-product operation (but warning: Euclidean spaces can have dimensions other than 3).

I'm thinking it should be a method on the `Vec` trait, requiring evidence that the dimension is indeed 3 (see #628 for fixed-size Vecs).

Another possibility: find a suitable generalization of the cross-product (exterior algebras?), and add a generic typeclass for that.

pleira:
Amazing the idea of using [exterior algebras](https://en.wikipedia.org/wiki/Exterior_algebra) for this. I did not know anything about that. It would be cool if the normal 3 dimensional cross-product would be a kind of corollary of a more generic algebra. Having said that, I am quite fine with a method implementation as #628 tries to do for fixed-size Vecs.

PR #628, opened 2017-02-22
Add support for vectors 

zainab-ali:
This is a POC for dimensional vectors.

It's currently using the shapeless `nat`, but I know we don't want this as a dependency.
Some options are:

- use singleton ops for singleton type operations.  I think this is compatable with 2.12 and 2.11, but I'm not sure about 2.10
- have objects representing natural numbers up to 4, since most people only care about 2, 3 and 4d vectors.  This is a bit of boilerplate, but could be autogenerated if necessary.

I'm also not sure about the name `Vec`.  I'd prefer `Vector`, but this would clash with the standard library.

What do you think?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -60,6 +60,7 @@ lazy val core = crossProject
   .settings(coreSettings:_*)
   .settings(buildInfoSettings:_*)
   .settings(crossVersionSharedSources:_*)
+  .settings(libraryDependencies += "com.chuusai" %% "shapeless" % shapelessVersion)
   .jvmSettings(commonJvmSettings:_*)
   .jsSettings(commonJsSettings:_*)
   .dependsOn(macros)
--- a/core/shared/src/main/scala/spire/math/Vec.scala
+++ b/core/shared/src/main/scala/spire/math/Vec.scala
@@ -1,16 +1,14 @@
 package spire
 package math
 
-import scala.{Vector => SVector}
-
+import scala.reflect.ClassTag
 import spire.algebra._
 import spire.implicits._
 
 import shapeless._
 import shapeless.ops.nat._
-import scala.reflect.ClassTag
 
-final class Vec[N <: Nat, A](val repr: SVector[A]) extends AnyVal {
+final class Vec[N <: Nat, A](val repr: Array[A]) {
   
   import LT._
   import LTEq._
@@ -22,20 +20,22 @@ final class Vec[N <: Nat, A](val repr: SVector[A]) extends AnyVal {
 
   def n = repr.size
 
-  def map[B](f: A => B): Vec[N, B] = new Vec(repr.map(f))
+  def map[B](f: A => B)(implicit act: ClassTag[A], bct: ClassTag[B]): Vec[N, B] = new Vec(repr.map(f))
 
-  def ap[B](ff: Vec[N, A => B]): Vec[N, B] = 
+  def ap[B](ff: Vec[N, A => B])(implicit act: ClassTag[A], bct: ClassTag[B]): Vec[N, B] = 
     new Vec(repr.zip(ff.repr).map { case (a, f) => f(a) })
 
-  def map2[B, C](that: Vec[N, B])(f: (A, B) => C): Vec[N, C] = 
+  def map2[B, C](that: Vec[N, B])(f: (A, B) => C)(implicit act: ClassTag[A], bct: ClassTag[B], cct: ClassTag[C]): Vec[N, C] = 
     that.ap(map(f.curried))
 
   def updated(i: Nat, a: A)(implicit ev: i.N < N,
-                           toInt: ToInt[i.N]): Vec[N, A] =
+                           toInt: ToInt[i.N],
+                           act: ClassTag[A]): Vec[N, A] =
     updated[i.N](a)
 
   def updated[NN <: Nat](a: A)(implicit ev: NN < N,
-                           toInt: ToInt[NN]): Vec[N, A] =
+                           toInt: ToInt[NN], 
+                           act: ClassTag[A]): Vec[N, A] =
     new Vec(repr.updated(toInt(), a))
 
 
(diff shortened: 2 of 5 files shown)


pleira:
I would like to see something like this in Spire. It would allow me to close #375: there is an implementation of cross product inside.

denisrosset:
While having fixed-size vectors is a good thing, I would make it available in an extra package.

Specifically:

1. one can reasonably want to use either compile-time fixed-size vectors (with type safety), or have runtime-sized vectors (when passing Nats around is inconvenient). So I prefer Spire to to be agnostic, especially with the blanket import "spire.math._".

2. Why having the whole `spire-core` depend on shapeless just for the sake of `Nat`? This can be solved by having fixed-size Vecs in their own package (the dependency on shapeless would be restricted there), or Vec using his own `Nat` implementation. 

3. There are outstanding issues in the Spire library, on which the core maintainers, myself included, make currently slow progress. I'm thinking of truncated division, sporadic bugs in the sparse polynomials, unlawful typeclasses in `spire.math`. I'd like to clean up what we currently have before extending the scope.

Apart from that, some suggestions:

1. why have `Vec` a value class that encapsulate the slow `scala.Vector`? Vector is not only slow because of the extra structure, but it will box its contents (i.e. a 4D Vec of doubles will allocate much more than the 4*8 bytes required to store the coefficients; probably 3x overhead minimum -- the header for each boxed `Double`, plus the pointers in the `Vector`; the overhead of `Vector` itself is probably O(1), but non-negligible at small sizes).

2. why not have `Vec` backed by a private array, which would be primitive for primitive types? You would either require a `ClassTag`, or use the approach described here https://pellucidanalytics.github.io/blog/optimistic-respecialization/

(Note: in my `scalin` library, I made the explicit choice of _not_ providing specialization in the base trait, due to my use case, exact arithmetic, that uses `SafeLong` or `Rational`. This is an opinionated decision, so a variety of spire-based vector implementations should coexist.)

3. why not have `Vec` a base specialized trait, and provide specific instances for 1,2,3 and 4D, where the coefficients are fields and not stored in a structure? The 1-4D instances can be private, and selected at construction time depending on the size.

denisrosset:
@pleira : I answered on your original PR.

zainab-ali:
I agree that  `Vec` should be backed by an `Array` with a `ClassTag`

After mulling over this, I think there are use cases for N dimensional vectors.  Hence, I would disagree with only having implementations for dimensions 1 to 4.  
There are several ways of supporting this:
1. Have a custom `Nat` implementation within `spire-core`.  This would be compatible with 2.11, but would mean adding a fair amount of code that arguably wouldn't belong in spire.
2. Use singleton types, along with `singleton-ops`.  This would only work with 2.12, or with 2.11 and the typelevel Scala compiler.   The vector code could be moved out into a separate module, so that `spire-core` is not dependent on `singleton-ops`.
3. Use shapeless, but move the code to a separate module, so that `spire-core` doesn't have that dependency.  This would be compatible with 2.11.

I'm leaning more towards moving this to a separate module, since `spire-core` contains core typeclasses, but few data types.  That module could be dependent on shapeless if 2.11 compatibility is desired, or singleton ops if not.

pleira:
I do welcome having Vec in a separate module because of the shapeless dependency in Scala 2.11. From Spire's User's Guide there could be a pointer to this Vec type so that it is found by the users. I mean, spire-core has Complex and Quaternion, and somehow, as a user, I would expect to find some type in between, like a Vec for the particular case of three dimensions. If is fine with me when Vec can provide types for higher dimensions as well.

denisrosset:
> Hence, I would disagree with only having implementations for dimensions 1 to 4.

Sorry for not being clear. I was thinking of:

```scala

trait Vec[@specializeed A, N <: Nat] {
  // declaration of inteface
}

// instances for Vec1 and Vec2 not shown
private[spire] class Vec3[@specialized A](x: A, y: A, z: A) extends Vec[A, _1] {
 // implementation
}

private[spire] class VecN[@specialized A, N <: Nat](els: Array[A]) extends Vec[A, N] {
 // implementation
}
```

denisrosset:
@pleira Think of it this way.

If you need to add new syntax to support a common algebraic structure, such as exterior algebras, create a new typeclass, enumerate the associated properties, create the corresponding law check definitions, add an enrichment implicit in `spire.syntax`.

I'm not saying we should add exterior algebras, or any algebraic structure we fancy however (let's keep the complexity in check!).

If you need to add new syntax to support a particular type (such as a 3D vector), think of adding the syntax directly on the type. It makes the operators easier to find.

zainab-ali:
I've been having something of a mental block on this one, mainly because I can't think of a decent way to support nats while minimising dependencies.

I'm also half thinking of moving this out to a separate library entirely, which could depend on Typelevel Scala and singleton ops.  

@denisrosset as a maintainer, what would you prefer?

- vectors in `spire-extras` with a dependency on shapeless
- vectors in `spire-extras` with a spire `Nat`
- vectors in a new `spire-vec` module with a dependency on Typelevel Scala and singleton-ops.  I'm not even sure if a single module can depend on Typelevel Scala.
-  have this in a different library entirely

denisrosset:
Literal singleton types seem (for me) the way to go, as they are available in Typelevel Scala (+DOTTY!) and are pushed by a SIP, reuse familiar terminology (singleton subtypes of Int), and won't, in the future, require dependencies.

I'd like to have standardized 2D and 3D lightweight vectors in Spire (however, there is a also space for a vector library with no compile-time size checks).

For now, I'd suggest you to write a separate library so that you can iterate faster --- and it's easier to open new issues to discuss the design, while discussion on a Spire branch is cumbersome.

And we can merge it back later, depending on the size in an existing module or in its own `spire-vec`.

denisrosset:
Hi @zainab-ali , we are cleaning up the PR/issue backlog, so this is moved to a specific issue.

Issue #598, opened 2016-11-03
What is exactly `Jet`?


denisrosset:
The status of `Jet` is unclear. There are conflicting indications that:

1. `Jet` is intended to be a substitute of a real number in computations, so that derivatives can be automatically recovered at the end. Thus `Jet` has `Field`, `Order` instances ... even if those instances are not lawful. In that case, the `Order` comparisons should be done on the non-infinitesimal part only.

2. `Jet` is an implementation of `n`-dimensional dual numbers, and is thus an algebra over the real numbers. In that case, no `Order` instance should exist, because dual numbers, as a `d > 1` vector space, do not have a total order.

denisrosset:
Ok.... there are a few issues.

- the dual numbers are a proper algebraic structure (see https://en.wikipedia.org/wiki/Dual_number ). Division is only defined when the real part of the denominator is nonzero. So they do not form a division ring. But they are a genuine ring, of course non ordered.

- you can use dual numbers in code that expects a real scalar; you then pretend to operate as a real number, while applying the differentiation rules. In that case, you want to pretend that `Jet[Real]` is ordered.

Jet also inherits the `ScalaNumeric` stuff, with ugly hacks to perform arithmetic on `Any`.

The implementation is however good and well-commented. It should be split into parts to correspond to the different functions described above.

PR #630, opened 2017-03-04
Scaled back the requirements of Complex/Jet from IsReal to Signed.

denisrosset:
Jet is still ambiguous, see #598


denisrosset:
This PR prepares the work on truncated division (with truncated division, we can have lawful `floor`, `ceil`, `round`).

IsReal is probably going to disappear (or be an alias for the intersection of lawful types).

denisrosset:
The patch coverage is misleading, check https://codecov.io/gh/non/spire/compare/46c1760c56eb5b242d30d14f3162553cc8dd2337...069b7ee3954171cc61d4b5641f2d6b279f641663

PR #633, opened 2017-03-10
Adding GCDRing laws, fixing implementations

denisrosset:
I fixed a few bugs, missing methods that reverted to the Euclidean algorithm... mainly around `lcm` support.

Rational.gcd is still failing the commutativity test.

Edit: Rational should be fixed now.

Edit: major change, `EuclideanRing` and `Field` do not provide default implementations of the GCD/LCM anymore, but these are available in traits in the companion.


core/shared/src/main/scala/spire/algebra/Field.scala
```diff
@@ -1,21 +1,61 @@
 package spire
 package algebra
 
+/** We precise AlgebraField[_] = algebra.Field[_] by extending GCDRing/EuclideanRing with
```
tixxit: I think this needs to be reworded - can't quite figure it out.

core/shared/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -539,6 +539,7 @@ trait PolynomialOverField[@sp(Double) C] extends PolynomialOverRing[C]
     }
   }
 
+  /* TODO: check if we can use directly Euclidean algorithm
   // TODO: why final?
   final override def gcd(x: Polynomial[C], y: Polynomial[C])(implicit ev: Eq[Polynomial[C]]): Polynomial[C] = {
```
tixxit: Should this commented code be deleted now?

core/shared/src/main/scala/spire/math/Rational.scala
```diff
@@ -58,10 +62,57 @@ sealed abstract class Rational extends ScalaNumber with ScalaNumericConversions
   }
    */
 
-  /* TODO: not commutative
-  def lcm(rhs: Rational): Rational = (lhs / (lhs gcd rhs)) * rhs
-  def gcd(rhs: Rational): Rational
-   */
+  def lcm(rhs: Rational): Rational = if (lhs.isZero || rhs.isZero) Rational.zero else (lhs / (lhs gcd rhs)) * rhs
```
tixxit: nit: space between these 2 methods - perhaps splitting lcm over a couple of lines.

core/shared/src/main/scala/spire/math/Rational.scala
```diff
@@ -58,10 +62,57 @@ sealed abstract class Rational extends ScalaNumber with ScalaNumericConversions
...
+  def gcd(rhs: Rational): Rational =
```
tixxit: Nice!

core/shared/src/main/scala/spire/math/Rational.scala
```diff
@@ -839,7 +856,14 @@ private[math] trait RationalApproximateNRoot extends NRoot[Rational] {
     Rational(n.toDouble ** k.toDouble)
 }
 
-private[math] trait RationalIsField extends Field[Rational] {
+private[math] trait RationalIsField extends Field.FieldOfFractionsGCD[Rational, SafeLong] {
+  override def gcd(x: Rational, y: Rational)(implicit ev: Eq[Rational]) = x.gcd(y)
```
tixxit: Since `gcd` and `lcm` are being overridden, is there any benefit in extending `FieldOfFractionsGCD` here?

core/shared/src/main/scala/spire/std/float.scala
```diff
@@ -6,7 +6,7 @@ import spire.math.Rational
 
 import java.lang.Math
 
-trait FloatIsField extends Field[Float] {
+trait FloatIsField extends Field.WithDefaultGCD[Float] {
```
tixxit: I think the existing gcd/lcm impl does some fancy stuff, but more matched the field of fractions impl instead. Any reason to prefer this one? I wrote the last one, so have some attachment, but definitely happy to see it go if it doesn't make sense anymore.

tests/src/test/scala/spire/laws/LawTests.scala
```diff
@@ -33,21 +33,21 @@ class LawTests extends FunSuite with Discipline {
   }
 
   // Float and Double fail these tests
-  checkAll("Int",        RingLaws[Int].euclideanRing)
-  checkAll("Long",       RingLaws[Long].euclideanRing)
+  checkAll("Int",        RingLaws[Int].cRing)
+  checkAll("Long",       RingLaws[Long].cRing)
```
tixxit: This, along with `Int` (and `Byte` + `Short`) can all be `signedGCDRing`, right? Should bump up the coverage a bit.

denisrosset: Two things (that should be documented):

- signedGCDRing is an add-on on top of other tests (so that I don't need to define signedEuclideanRing, signedField, orderedGCDRing, orderedEuclideanRing, ...)

- the limited range types cannot be tested with Signed, due to wrap on overflow, see the next PR #634 for a solution (let us discuss it there)

core/shared/src/main/scala/spire/std/float.scala
denisrosset: Cool! I had no idea of the laws implemented by the fancy stuff.

Do you remember what the commented out code computed?

If it is approximately lawful, I'll be happy to switch. There are two ways to "prove" that it is approximately lawful.

1) either document that the operation respects lcm(x, y) * gcd(x, y) = x y + commutativity + associativity 

2) or say that the result should match closely the result obtained with Rational; and we can even have exact tests by restricting our Double/Floats to fractions with denominator 2^n

core/shared/src/main/scala/spire/math/Rational.scala
denisrosset: No. I corrected this in a later PR. Let me bring the changes here.

core/shared/src/main/scala/spire/math/Polynomial.scala
denisrosset: Arg. I wanted to look at Polynomial later; let me check.

core/shared/src/main/scala/spire/algebra/Field.scala
denisrosset: OK, it means that algebra.Field only extends CRing, and algebra does not define the intermediate types.

We define our own Field to have the inheritance, and provide good defaults for gcd/lcm/quot ...

core/shared/src/main/scala/spire/std/float.scala
tixxit: Yeah - I think 2) may be an OK way to go. Specifically, I think the error on the final GCD is actually relative to the result +/- eps or something very near. If we merge this with the code commented out, I can make a PR to add a test that re-enables the old behaviour, but adds a test that it's approximately equal to the `Rational` case w/ 2^n denom.

tests/src/test/scala/spire/laws/LawTests.scala
tixxit: SG

core/shared/src/main/scala/spire/algebra/Field.scala
denisrosset: Is corrected in the next commit.

core/shared/src/main/scala/spire/math/Polynomial.scala
denisrosset: OK... the trick is that Polynomial.gcd had a special test to return a nice number in case the gcd was a constant.

As the gcd is defined up to a unit, we could keep this postprocessing, maybe extend it to non-constant results. Keeping only the special case feels wrong; let's rediscuss the matter later #641

denisrosset: Removing the commented out code for now.

core/shared/src/main/scala/spire/math/Rational.scala
denisrosset: Done

core/shared/src/main/scala/spire/std/float.scala
denisrosset: OK cool!

denisrosset: It should actually be exactly equal; when the denom is 2^n, the operations are exact, as long as they don't overflow the mantissa.

tixxit:
This all looks great! I think everything has been either fixed, addressed, or put in an issue. Feel free to merge whenever!

PR #634, opened 2017-03-12
Laws for limited range primitive types. Bug fixes for Byte and Short.

denisrosset:
Based on #633, which should be merged first.


denisrosset:
OK... changed the strategy from rejecting generated values to shrinking the generated values.

This avoids failing tests, at the price of a syntax addon.

PR #639, opened 2017-03-15
Create FRIENDS.md listing Spire-friendly libraries

denisrosset:


clayrat:
Seems like the `sonicreducer` lib is only using Spire in tests, when they reduce an array of rationals.

PR #640, opened 2017-03-19
Add basic release notes for 0.14.0.

non:
This primarily updates the version number used in the README, and adds (very
terse) release notes for 0.12, 0.13, and 0.14.

We still need to update the GUIDE and README with new types and changes, but
this is a start.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,58 @@
+## Version 0.14.0
+
+This release is a major milestone for Spire.
+
+First of all, it introduces a dependency on typelevel/algebra (and
+thus on typelevel/cats-kernel). This provides immediate compatibility
+with Cats and Algebird.
+
+Second of all, Spire is now published to the org.typelevel
+organization (which was previously org.spire-math).
+
+Additions:
+ * Add GCDRing, generalized from EuclideanRing
+ * Add DivisionRing, a field without commutative multiplication
+
+Changes:
+ * Many type classes are now aliases to algebra's type classes
+ * The relationship between EuclideanRing and Gcd has changed
+ * IsReal is no longer required to be ordered, only signed
+ * Many different efficiency improvements
+
+Fixes:
+ * Instances for Complex and Quaternion are more precise
+ * Numeric[Complex[A]] is no longer provided
+ * Many bug fixes
+
+## Version 0.13.0
+
+Additions:
+ * Added Eq[Bound[A]] instance
+ * Added Interval#overlap and supporting machinery
+
+Changes:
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,58 @@
...
+Second of all, Spire is now published to the org.typelevel
+organization (which was previously org.spire-math).
+
+Additions:
+ * Add GCDRing, generalized from EuclideanRing
+ * Add DivisionRing, a field without commutative multiplication
+
+Changes:
+ * Many type classes are now aliases to algebra's type classes
+ * The relationship between EuclideanRing and Gcd has changed
```
denisrosset: Compared to algebra `Field <: CRing`, we have the hierarchy `spire.algebra.Field <: EuclideanRing <: GCDRing <: CRing`, with `spire.algebra.Field` extending both `EuclideanRing` and `algebra.Field`.

I was waiting for the @tixxit microsite to document these changes; if we release, I should update the original GUIDE as well.

CHANGES.md
```diff
@@ -1,3 +1,58 @@
...
+ * IsReal is no longer required to be ordered, only signed
```
denisrosset: To be precise: Signed now extends Order.

Also, Signed should be interpreted as a linearly ordered group, but we keep two typeclasses "type towers" to avoid proliferation of types/ambiguous implicits (also, Sign has a Signed instance without being a additive monoid).

However, we check the laws of linearly ordered groups.

CHANGES.md
denisrosset: We should also mention that GCDRing and EuclideanRing operations (gcd, lcm, quot, mod, quotmot) have been made lawful for all types.

We now require `gcd` and `lcm` commutative.

For integers, the result of the above operations is unchanged (apart fixing the corner cases of gcd/lcm).

But Rational, Double, Float... implemented originally EuclideanRing with truncated division, which is incorrect and will affect e.g. polynomial rings as well. This has been corrected. In some cases, it impacts gcd/lcm as well.

Trouble now: Java/Scala do not represent Euclidean division for `/~` `%` and `/%`, rather truncated division. So our syntaxes diverge there (this affects only stdlib Float/Double, which is why the issue never popped up in tests).

non:
@denisrosset Feel free to push your changes to `topic/release-notes-0.14.0` -- I'm sure whatever you write will be fine.

non:
I'm going to merge this PR so we can make the release announcement. Thanks @denisrosset!

denisrosset:
Perfect, thanks!

PR #642, opened 2017-03-27
[READY] Tests of primitive types with limited range

denisrosset:
Supercedes #634, based on new master (PR easier to read)

This implements law tests for types that approximate an arbitrary range type.

For example, Byte/Short/Int/Long approximate a BigInt, and their operations are defined as long as the range of the primitive type is not exceeded. This PR adds specialized versions of the law tests that verify that the randomized checking occurs within the range of the primitive type.


denisrosset:
Renaming as [WIP], because this should be addressed during the revamping of the laws suite (migration to newer cats 1.0 style).

denisrosset:
Ready for review.

denisrosset:
Closing for now.

PR #651, opened 2017-10-09
[READY] Moving basic performance utils into spire-kernel package

denisrosset:
We create a spire-kernel package including:

- the `spire.util.Opt` option-like wrapper,
- the `cfor` family of macros.


denisrosset:
Done, review by @tixxit, @non or @rklaehn .

PR #653, opened 2017-10-09
[READY] Identification and correction of the (transient) bug in sparse polynomial handling.

denisrosset:
The bug appeared when constructing a sparse polynomial from a list of Term, and a pair of terms of the form `(-a, n)` and `(a, n)` was present, corresponding to -a x^n + a x^n = 0. The resulting coefficient = 0 was not filtered out and led to bugs elsewhere, as an invariant of both sparse and dense polynomial is that no coefficient present should be equal to 0.

Note: I had to pump ScalaCheck `minSucessful` to 5000 for the bug to be reproducible.


ryan-williams:
Broken tests that this fixes failed several of my builds yesterday and today, then I didn't see this and attempted my own fix at #671 😔

Would be good to get this merged!

PR #666, opened 2018-02-14
fix div by zero in Complex.norm

### LINKED ISSUE #661: Complex(0.0, 0.0).norm == NaN
I expected a result of `0` here:

```scala
import spire._, math._, implicits._, algebra._
Complex(0.0, 0.0).norm  // NaN
```

[This is the cause in 0.14.1](https://github.com/non/spire/blob/v0.14.1/core/shared/src/main/scala/spire/math/package.scala#L408):

```scala
final def hypot[@sp(Float, Double) A](x: A, y: A)
  (implicit f: Field[A], n: NRoot[A], s: Signed[A]): A = {
  import spire.implicits._
  def abs(n: A): A = if (n < f.zero) -n else n
  if (x > y) abs(x) * (1 + (y/x)**2).sqrt
  else abs(y) * (1 + (x/y)**2).sqrt
}
```

Seems like a convoluted way to do this computation, is there a reason to not just say e.g. `(x*x + y*y).sqrt?

I love this library btw 😍  🙏🏻

ryan-williams:
fixes #661

### DIFF
--- a/core/shared/src/main/scala/spire/math/Complex.scala
+++ b/core/shared/src/main/scala/spire/math/Complex.scala
@@ -98,8 +98,7 @@ final case class Complex[@sp(Float, Double) T](real: T, imag: T)
   def complexSignum(implicit f: Field[T], n: NRoot[T], s: Signed[T]): Complex[T] =
     if (isZero) this else this / abs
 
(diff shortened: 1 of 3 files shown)


tixxit:
LGTM

PR #667, opened 2018-02-15
fix Complex.{abs,norm} for negative reals

ryan-williams:
fix a bug introduced by #666: `abs` and `norm` returned negative values when one of {real,imag} was zero and the other was negative

there was discussion on #661 around whether `hypot` should assume non-negative arguments and `Complex.{norm,abs}` should correct for them before calling `hypot`, but this felt simpler and relatively downside-free

### DIFF
--- a/core/shared/src/main/scala/spire/math/package.scala
+++ b/core/shared/src/main/scala/spire/math/package.scala
@@ -404,10 +404,11 @@ package object math {
     (implicit f: Field[A], n: NRoot[A], s: Signed[A]): A = {
     import spire.implicits._
     def abs(n: A): A = if (n < f.zero) -n else n
-    if (x == f.zero) abs(y)
-    else if (y == f.zero) abs(x)
-    else if (x > y) abs(x) * (1 + (y/x)**2).sqrt
-    else abs(y) * (1 + (x/y)**2).sqrt
+    val (ax, ay) = (abs(x), abs(y))
+    if (x == f.zero) ay
+    else if (y == f.zero) ax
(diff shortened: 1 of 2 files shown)


ryan-williams:
coverage decrease seems spurious to me, dunno if that's normal / common?

core/shared/src/main/scala/spire/math/package.scala
```diff
@@ -404,10 +404,11 @@ package object math {
     (implicit f: Field[A], n: NRoot[A], s: Signed[A]): A = {
     import spire.implicits._
     def abs(n: A): A = if (n < f.zero) -n else n
-    if (x == f.zero) y
-    else if (y == f.zero) x
-    else if (x > y) abs(x) * (1 + (y/x)**2).sqrt
-    else abs(y) * (1 + (x/y)**2).sqrt
+    val (ax, ay) = (abs(x), abs(y))
```
denisrosset: Is the `Tuple2` allocation/pattern matching optimized away?

I'd prefer to see `val ax = abs(x); val ay = abs(y)`, especially in specialized methods.

ryan-williams: good point, I wouldn't be surprised if it's not, just fixed

denisrosset:
@ryan-williams Thanks for helping! The coverage fluctuations come from the randomized property tests, not a big deal.

denisrosset:
I'll merge after it passes C.I.

ryan-williams:
makes sense re: coverage, and thanks for reviewing and helping with this library!

PR #668, opened 2018-02-15
fix infinite loop in BigDecimal.sqrt

### LINKED ISSUE #664: Infinite loop in BigDecimal.sqrt
```scala
import spire.implicits._
BigDecimal("4.000000000000000000000000000000003").sqrt
```

enters an infinite loop [here](https://github.com/non/spire/blob/v0.14.1/core/shared/src/main/scala/spire/std/bigDecimal.scala#L52):

```scala
@tailrec def loop(x: BigDecimal, y: BigDecimal): BigDecimal =
  if (x == y) y else loop(y, ((n / y) + y) / two)
```

with values:

```
x: 2.000000000000000000000000000000001
y: 2.000000000000000000000000000000000
n: 4.000000000000000000000000000000003
```

affects `0.13.0` and `0.14.1`

ryan-williams:
fixes #664

### DIFF
--- a/tests/src/test/scala/spire/math/PackageTest.scala
+++ b/tests/src/test/scala/spire/math/PackageTest.scala
@@ -25,11 +25,11 @@ class PackageTest extends FunSuite {
     // exact values.
     assert(pow(BigDecimal(2), BigDecimal(5)) === BigDecimal(32))
 
-    // wolfram alpha says
-    // 485165146.8932734862594040580748685386577025894303647601088074911328839160579383035755177941692746135770688529810692...
-
+    // wolfram alpha says:                                                         485165146.8932734862594040580748685
     assert(pow(BigDecimal("1.00000001"), BigDecimal("2000000000")) === BigDecimal("485165146.8932734862594042197965646"))
-    assert(pow(BigDecimal(Double.MaxValue) * 100, BigDecimal("0.00001")) === BigDecimal("1.007169457206495657176075115660221"))
+
+    // wolfram alpha says:                                                               1.007169457206495657176120750097341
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/spire/math/PackageTest.scala
```diff
@@ -25,11 +25,11 @@ class PackageTest extends FunSuite {
...
 
-    // wolfram alpha says
-    // 485165146.8932734862594040580748685386577025894303647601088074911328839160579383035755177941692746135770688529810692...
-
+    // wolfram alpha says:                                                         485165146.8932734862594040580748685
     assert(pow(BigDecimal("1.00000001"), BigDecimal("2000000000")) === BigDecimal("485165146.8932734862594042197965646"))
-    assert(pow(BigDecimal(Double.MaxValue) * 100, BigDecimal("0.00001")) === BigDecimal("1.007169457206495657176075115660221"))
+
+    // wolfram alpha says:                                                               1.007169457206495657176120750097341
+    assert(pow(BigDecimal(Double.MaxValue) * 100, BigDecimal("0.00001")) === BigDecimal("1.007169457206495657262139152506902"))
```
ryan-williams: NB: this is now closer to the Wolfram Alpha value than it was previously

tixxit:
This has a conflict from merging the `PolySparse` fix - once the conflict is resolved, will merge!

ryan-williams:
thanks @tixxit! on it, likewise #669

ryan-williams:
the above was an attempt at using github conflict-resolution UI, which I think will work, but I was otherwise going to push a rebased version that removes my bad+redundant attempt at #653 from the commit history, I'll force-push that momentarily and we can let the tests run on it

ryan-williams:
all set here @tixxit when you have a minute, thanks!

lmk if you'd prefer that I follow the `[READY]` PR-title convention I saw in #653

PR #669, opened 2018-02-15
fix Complex.sqrt, add tests

### LINKED ISSUE #662: Complex[Double].sqrt returns wrong answers
```scala
import spire.math.Complex, spire.implicits._

val cs = 
  Seq(
    Complex( 3.0,  4.0), 
    Complex( 3.0, -4.0), 
    Complex(-3.0,  4.0), 
    Complex(-3.0, -4.0)
  )

println(cs.map(_.sqrt).mkString("\n"))
// (2.0 + 1.0i)
// (2.0 + -1.0i)
// (2.0 + 1.0i)   wrong!
// (2.0 + -1.0i)  wrong!

println(cs.map(_.nroot(2)).mkString("\n"))
// (2.0 + 1.0i)
// (2.0 + -1.0i)
// (1.0000000000000002 + 2.0i)   // right
// (1.0000000000000002 + -2.0i)  // right 
```

Affects 0.13.0 and (`org:typelevel`) 0.14.1

ryan-williams:
fixes #662 

first two commits are just #667 

can merge this one instead of #667, or do #667 first then merge this one on top of that, whatever maintainers want

### DIFF
--- a/core/shared/src/main/scala/spire/math/Complex.scala
+++ b/core/shared/src/main/scala/spire/math/Complex.scala
@@ -222,20 +222,22 @@ final case class Complex[@sp(Float, Double) T](real: T, imag: T)
   // returns the root with angle in (-pi/2, pi/2]
   def sqrt(implicit f: Field[T], n0: NRoot[T], s: Signed[T]): Complex[T] = {
     if (isZero) {
-      Complex.zero[T]
+      this
+    } else if (imag.isSignZero()) {
+      if (real.isSignNegative())
+        Complex(f.zero, real.abs.sqrt)
+      else
+        Complex(real.abs.sqrt, f.zero)
     } else {
+      // https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Negative_or_complex_square
       val two = f.fromInt(2)
+      val abs = this.abs
       val a = ((abs + real) / two).sqrt
-      imag.signum match {
-        case 0 =>
-          if (real < f.zero)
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/spire/math/Complex.scala
```diff
@@ -219,17 +219,21 @@ final case class Complex[@sp(Float, Double) T](real: T, imag: T)
...
   }
 
+  // returns the root with angle in (-pi/2, pi/2]
   def sqrt(implicit f: Field[T], n0: NRoot[T], s: Signed[T]): Complex[T] = {
     if (isZero) {
       Complex.zero[T]
     } else {
       val two = f.fromInt(2)
-      val a = ((abs + real.abs) / two).sqrt
+      val a = ((abs + real) / two).sqrt
```
denisrosset: Why should `a` be computed outside the match?

The whole function can be replaced by the following pseudo-code:

- `imag.isSignZero` and `real.isSignNegative`
  Return `Complex(f.zero, (-real).sqrt)`
- `image.isSignZero` and `real.isSignPositive` 
  Return `Complex(real.sqrt, f.zero)`
- Otherwise (add reference to [Wikipedia](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Negative_or_complex_square) )
  Compute and store `this.abs` in a temporary variable

The correction is fine.

denisrosset:
Seems good!

PR #672, opened 2018-02-17
migrate sbt 1.x

xuwei-k:


denisrosset:
@xuwei-k thanks! As I'm not the original author, going to way a few days to see if @non or @tixxit have a comment. Then I'll merge.

.travis.yml
```diff
@@ -3,14 +3,10 @@ language: scala
 scala:
   - 2.10.6
   - 2.11.8
+  - 2.12.0
```
ryan-williams: is the policy to build+test against the earliest patch-version for each minor-version, to ensure maximum compatibility?

xuwei-k: I have just copied from 

```yaml
matrix:
  include:
  - scala: 2.12.0
    jdk: oraclejdk8
```

PR #680, opened 2018-03-19
Add a docs project using tut.

non:
The goal here is to type-check our docs to avoid code rot. We
desperately need more docs and this is a way to help ensure we keep
things up-to-date.

Eventually we'd like to use sbt-microsites. Right now these aren't
compatible with sbt 1.0, so we're going to produce markdown
documentation here and figure that we'll eventually be able to put it
*somewhere*.

Apologies to Tom: this probably supersedes the old microsites PR.


non:
Confirmed that tests pass locally. Tired of waiting for Travis.

PR #686, opened 2018-03-19
Modularization Part 1

non:
This PR is the first step to a modular Spire.

We split out `platform`, `util`, and `data` (currently empty). Future PRs will further modularize.

Highlights:

 - drop 2.10 support
 - `core` has no `shared` directory any more
 - remove paradise dep
 - remove scala_2.10, etc.

Review by @denisrosset


non:
Looks good to me. 👍 

In the long run we're going to improve coverage but I'm fine with this for now.

Apologies to everyone whose going to lose branches/PRs due to this change. I think in the long run this restructuring will be very important.

PR #692, opened 2018-03-20
Check README.md with tut

non:
We had a few errors in the README which this catches and fixes.

In the long run I'd like to figure out how to run `tut` on README.md directly, but at least this helps us check things, with a bit of manual copying.


denisrosset:
LGTM.

PR #693, opened 2018-03-20
Truncated division redux

denisrosset:


denisrosset:
@non is ok we are drinking beers

PR #698, opened 2018-03-20
Return BigDecimal values with better math contexts.

non:
This is a port of #584. On that PR, Tom said:

> Since we weren't explicitly setting the MathContext, it was using
> the default. However, by default, BigDecimal won't round the input,
> so it looked OK. However, any operations performed with this number
> as the receiver of the method would've truncated the
> result. Probably unexpectedly. Ugh... Scala's BigDecimal is such a
> mess.

This replaces https://github.com/non/spire/pull/584.


denisrosset:
LGTM

PR #659, opened 2017-11-28
Differentiate Left and Right Modules

stephen-lazaro:
Resolves https://github.com/non/spire/issues/656.
I noticed there are no laws for module structures. Would adding them be wanted?

PR #701, opened 2018-03-20
Left and right modules

denisrosset:
Contains the fixes of #659, thanks to @stephen-lazaro


denisrosset:
Discussed with @non around beers.

PR #704, opened 2018-03-22
Tests for limited range types

denisrosset:
We add shadows to the tests: during testing, a primitive type is shadowed by a unlimited range type. When any intermediate computation exceeds the range of the primitive type, an exception is thrown such that the property evaluation is discarded (for now, we force it to pass to avoid failing on too many discards).

Changes:

- Introduces `spire.laws.shadows` which provides a `Shadow[A, S]` type containing a value `a: A` and its unlimited range `s: S`. `S` will usually be `SafeLong` or `BigInt`.
- Introduces wrappers for Spire typeclasses in `spire.laws.shadows`.
- Introduces `InvalidTestException`, thrown when 
- Introduces a wrapper for the laws, such that this exception leads to properties passing or being undecided.

The design is not final. In particular, I'm looking at:

- Improving the syntax of the wrappers that catch the `InvalidTestException`. I will probably replace the scalacheck `forAll` method, as this would have the minimal impact on the law syntax.

- Should we force the property to pass when the range is exceeded, or send `undecided`? `undecided` has the unfortunate effect to make some properties very difficult to pass, especially for `Byte`-sized types.

- Grouping the files in `spire.laws.shadows` which currently follow cats testing style.

- Add range-proof tests for all primitive types and all laws (as is done currently in the experimental branch).

This would improve a lot the coverage for primitive types (I get +10% on the unsigned primitive types).

### DIFF
--- a/laws/src/main/scala/spire/laws/CombinationLaws.scala
+++ b/laws/src/main/scala/spire/laws/CombinationLaws.scala
@@ -26,10 +26,10 @@ trait CombinationLaws[A] extends BaseLaws[A] {
     name = "signedAdditiveCMonoid",
     parent = None,
     "ordered group" → forAll { (x: A, y: A, z: A) =>
-      (x <= y) ==> (x + z <= y + z)
+      checkTrue(!(x <= y) || (x + z <= y + z)) // replaces ==>
     },
     "triangle inequality" → forAll { (x: A, y: A) =>
-      (x + y).abs <= x.abs + y.abs
+      checkTrue((x + y).abs <= x.abs + y.abs)
     }
   )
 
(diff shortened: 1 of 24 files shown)


denisrosset:
Test coverage is ironically decreased because only a subset of tests are currently executed.

non:
@denisrosset One idea I had was to make the behavior when the range is exceeded configurable.

I think defaulting to *pass* is appropriate. Other options I could imagine are:
 - *undecided* has the downsides you mentioned
 - we could include a *transformation* to map larger values back into the space

Since this isn't C, I don't like the idea of considering range-exceeding behavior undefined in cases where it's fairly well-understood (i.e. standard signed integer addition).

denisrosset:
Primitive types wrap, which is well defined under the ring laws.

However, the behavior is only compliant wrt Order/GCD... when we stay in range.

Alternative: include law variants for types that wrap when necessary.

non:
@denisrosset Ah, I see, OK.

Anyway, I'm 👍 on this as long as it doesn't cause our properties to become too flaky (defaulting to *passing* should ensure this doesn't happen).

Issue #718, opened 2018-03-23
Add TruncatedDivision instance for UInt, UShort, UByte


denisrosset:
A TruncatedDivision instance of ULong is part of a PR about limited range testing. It should be added to other unsigned types.

denisrosset:
Done in #719.

PR #719, opened 2018-03-23
More limited range tests / better syntax

denisrosset:
Continuing the work of #704 with simplified syntax, and more types covered.

The simplified syntax is a new `forAllSafe` method that wraps the `InvalidTestException` as a property that passes.

Remains:

- [ ] Add TruncatedDivision to more types #718 
- [ ] Document the shadowing machinery

(@non: the PR is a follow-up, no deep inspection is needed. If you happen to have a look, merge immediately, as the new syntax is an order of magnitude better than the previous PR we merged)


denisrosset:
Pinging @non: can you quickly review this? The syntax is *much* better than before, and coverage improves by a whole percent!

Note: at the end of the week, I'll merge anyway, as I don't see anything controversial in that PR.

PR #720, opened 2018-04-06
Get rid of Order context bound in Interval constructors

vpavkin:
Pushes `Order` requirements down to individual methods of `Interval`.

I didn't touch `IntervalSet`s though. @rklaehn do you have anything in mind that can be improved now?


denisrosset:
Looks fine to me. Merging, even if IntervalSet has still an internal instance.

PR #723, opened 2018-04-27
Involution typeclass. Orthogonal reqs for Complex/Quaternion.

denisrosset:
Added a Involution typeclass, with instances for Complex/Quaternion,
and trivial instances for types having IsReal instances.

Made Complex require (and implement) commutative instances. Made the
Complex requirements stricter, as Complex operations now require at
least a CRing.

Made the typeclasses derived for Complex[A] more regular (an underlying CRing gives you a RingAssociativeAlgebra for example).

Made Quaternion require CRing as well.


core/src/main/scala/spire/algebra/Conjugation.scala
```diff
@@ -0,0 +1,27 @@
+package spire
+package algebra
+
+/** Describes a conjugation, which is an antihomomorphism. It obeys:
+  *
+  *   i. conjugate(1) = 1
+  *  ii. conjugate(x + y) = conjugate(x) + conjugate(y)
+  * iii. conjugate(x*y) = conjugate(y)*conjugate(x)
```
non: I'm pretty sure `conjugate(conjugate(x)) = x` right? Can we add that here (we are testing it).

non:
This looks good, I just had one comment about a missing law in the documentation. Otherwise 👍 from me.

denisrosset:
@non: thinking of it, we should not call this typeclass `Conjugation`, as conjugation from the conjugate elements of field extensions. Our typeclass only really applies to the extension of real numbers to complex numbers, where each number has only one conjugate element. Other extensions are not so (for example cyclotomic numbers over rational numbers).

What we are looking is something that would apply as well to other rings, such as complex matrices. It's closer to [Involutive rings](https://en.wikipedia.org/wiki/*-algebra). Indeed, the missing law you mention should be present.

I'm going to propose a rename of the typeclass to `Involution`, with the `conjugate` operation named `star`.

BTW, we discussed splitting unprincipled stuff into the `legacy` package before deprecating it. We should have the reverse: an incubator package where stuff would be promoted to `core` after we got sufficient feedback. The package `extras` could play that role.

non:
I like using `extras` as an incubator, although for stuff we want to eventually delete I'm not sure I'd like to put it back in `extras` (it doesn't communicate legacy/deprecation).

The `Involution` name change is good, but I'm a bit wary of using `star`. I've been using Kleene algebras in a few places and they also have a `star` operation, which has different laws. Could we perhaps call it `adjoin` or something? Any other ideas?

denisrosset:
OK. I'm good to go for `adjoint`.

non:
👍

denisrosset:
Waiting for the CI to finish and will merge afterwards.

core/src/main/scala/spire/algebra/Conjugation.scala
denisrosset: Done in the new version.

adampingel:
@non Was it the verb `adjoin` or the adjective/noun `adjoint` that you intended?

PR #724, opened 2018-04-27
Added new packages as dependencies for the root project.

denisrosset:
The packages data, legacy, platform and util were not published due to them not being included in the root project.


denisrosset:
@non: please +1, this follows our work last month

non:
👍

PR #726, opened 2018-05-01
use foldLeft instead of /:

xuwei-k:
https://github.com/scala/scala/blob/5661290bfda0a08a60fb8df3ac60baafb55c0306/src/library/scala/collection/IterableOnce.scala#L465


denisrosset:
Thanks for that nice fix!

PR #752, opened 2018-12-15
Syntax test

yerar:
Added some test for Unicode optional operators on Intervals.


denisrosset:
@yerar: I've updated this PR, see #762. See the comment there about authorship.

Issue #742, opened 2018-10-16
publish for Scala 2.13.0-M5?


SethTisue:

erikerlandson:
@SethTisue +1
it is like you read my mind

erikerlandson:
`scalacheck-1.14.0` and `scalatest-3.0.6-SNAP3` are now available for 2.13.0-M5

`machinist` and `algebra` are outstanding

denisrosset:
Seems we are waiting for https://github.com/typelevel/algebra/issues/219

SethTisue:
algebra 1.0.1 is now available for M5

stijndehaes:
I think this ticket can be closed since PR is merged?

larsrh:
It is merged, but not published yet.

denisrosset:
@larsrh I'm travelling as well until Thursday! Would be happy to share the publishing process with you, if we can perform one release together to see how it works.

larsrh:
@denisrosset Sorry, I missed your comment. I'm running a release right now.

For the future, please create a Sonatype account and request publishing rights here: https://issues.sonatype.org/browse/OSSRH-4953

Let me know once you have done so and I will confirm your membership.

larsrh:
0.16.1 is out.

PR #759, opened 2019-02-03
Support cross-compiling build with 2.13.0-M5

erikerlandson:
PR for #742


erikerlandson:
note to self, use https://github.com/scala/scala-collection-compat ?

erikerlandson:
commit above cross compiles for main src, but test code compile fails

larsrh:
I don't have a strong opinion, but I'd prefer if you could use scala-collection-compat.

erikerlandson:
@larsrh I am using it.  OTOH I might not be using it in the best possible way, since I'm figuring out both the spire code and compat somewhat by trial and error.

larsrh:
> I am using it.

Sorry, I must've missed that. Unfortunately I currently don't have the bandwidth to aid in the process.

denisrosset:
I've had a look at your ongoing work, thank you so much! It can be quite painful to move a beast of a codebase like Spire.

Some of the things you've been modifying are quite obscure, such as the partial algebras (groupoids and the like). Happy to cross check your progress.

I suggest the following: let's not try to upgrade everything at the same time. A source of breakage is the immtuable `Seq` in 2.13; let's keep it for later.

Instead, we should stay on 2.11/2.12 and upgrade the collection syntax (factories, etc...), merge that. Then, in a second step, address the immutable Seq.

(I haven't checked, though, whether `spire-collection-compat` renders `Seq` immutable)

erikerlandson:
Thanks @denisrosset!  

I have been fooling around with getting test code to compile for 2.11 (which is likely to also work for 2.12 when it does).

You can see that the bulk of the breakage (no surprise) has been around the removal of TraversableLike, and CanBuildFrom -> Factory.  I got somewhat rat-holed drilling into why the testing code is failing to resolve the new implicit signatures, although I may have just needed `include scala.collection.compat._` in some additional places.

One thing I'd love to find is a write-up of what cross-compiling code is "supposed" to look like. "If you are doing everything right, your code should look analogous to these examples..."

Based on source diving, I think it's supposed to be "use Iterable and friends as the replacement for Traversable, and use Factory as the replacement CBF" but until I can compile the testing my confidence isn't very high 😁

denisrosset:
You know what? After you get part of this done, let's write a post for the Typelevel blog. That would be greatly valuable information.

(And you know more than me about the transition at this point!)

Very happy to have `Iterable` as a replacement for `Traversable`, and `IterableOnce` (which is new in compat/2.13) as a replacement for `TraversableOnce`.

larsrh:
#765 is merged now which contained the bumped dependency part of this PR.

erikerlandson:
just a note on current state.  the analog of `SeqLike` in 2.13 is `SeqOps` which has two interesting consequences.
1. spire also defines a thing called `SeqOps`, which is more of an annoyance than anything else but is obviously a bit confusing.
2. In the spirit of cross-compiling but also attempting to put a foot forward into the future, I defined a `spire.scalacompat.SeqOps` so that pre-2.13, it looks like `trait SeqOps[A, CC[_], C] extends SeqLike[A, CC[A]]`. This was promising but it caused some re-definitions of things like `SA` to `CC[A]` which "worked" until it started to break `@sp` for some reason, I am assuming something to do with  `@sp` being unable to resolve the entire sequence type `CC[A]` with these new signatures.

Next experiment will be defining something like `spire.scalacompat.MySeqLike[A, Repr]` that somehow can resolve to the 2.13 SeqOps.  scala 2.13 itself defines this (deprecated) as `SeqOps[A, Seq, Repr]` - the `Seq` here seems unappealing but maybe it's not as bad as I think.

The general theme with this has been that spire leverages pre-2.13 constructs (SeqLike, Traversable, @specialized, etc) in such a highly coupled way that cross-compiling to 2.13 is tricky

erikerlandson:
Another thing I'm going to try is replacing signatures of the form `SA <: SeqLike[A, SA]` with `SA <: SeqOps[A, Seq, SA]`, possibly that will avoid the cases where I broke `@sp`

denisrosset:
`@sp` is notably fickle.

And even when it compiles, it sometimes still maps to nonspecialized code, so buyer beware.

For these collection enrichment operations, why not fork the pre-2.13 and 2.13 code? and maintain both versions?

Also, let's rename `SeqOps` and `IndexedSeqOps` to `RichSeqOps` and `RichIndexedSeqOps` in a separate PR, with a comment explaining the deviation to the rule `XXX` + `Ops`.

erikerlandson:
I started over with a fresh branch (not pushed yet) that is somewhat cleaner. The `Factory` and `SeqOps` stuff I have compiles and tests for pre-2.13, but is snagging for 2.13 on some of the `SeqOps[A, Seq, SA]` on account of binding `CC[_]` elsewhere, which it of course cannot prove is `<= Seq[_]`.  I have other ways to square this circle, but it was worth a try.

erikerlandson:
Latest push cross-compiles for the libraries (with the exception of `simplification.scala` example), but I am hitting some diverging implicit resolutions when I try to build the tests on 2.13

erikerlandson:
Hmm, JS testing doesn't work. I only ran the JVM testing.
```
[info] Fast optimizing /home/travis/build/typelevel/spire/tests/.js/target/scala-2.11/spire-tests-test-fastopt.js
[error] Referring to non-existent class scala.collection.compat.package$
```

erikerlandson:
@SethTisue, does the above JS failure make any sense to you? It doesn't appear to be a problem with spire's `build.sbt` deps.

erikerlandson:
The latest also builds and tests on 2.13. I had to comment out a few tests, which I haven't diagnosed yet.

examples/src/main/scala/spire/example/simplification.scala
```diff
@@ -1,12 +1,19 @@
 package spire
 package example
 
+object Simplification {
```
larsrh: Some leftover code here?

erikerlandson: this example doesn't compile right now, due to incompatability around `newBuilder` method - you *have* to override pre-2.13, but post 2.13 the `override` is a compile errror

SethTisue:
>  "org.scala-lang.modules" %% "scala-collection-compat" % "0.3.0"))

@erikerlandson should be triple-percent?

erikerlandson:
lol
```
[error] /home/travis/build/typelevel/spire/compat/src/main/scala-2.13/package.scala:11:4: Regular expression matched 'println'
[error] /home/travis/build/typelevel/spire/compat/src/main/scala-2.13/scalacompat.scala: File must end with newline character
```

erikerlandson:
> @erikerlandson should be triple-percent?

@SethTisue that was it, thanks!

examples/src/main/scala/spire/example/simplification.scala
denisrosset: I don't see why this example needs to connect with the scala collection hierarchy. It's being used internally in one of the example cases; so let's just copy/paste the collection methods that are required for the example to work and move on.

(I don't understand what ties the enumeration functions and the simplification part; and I don't recognize the simplification algorithm. Does it correspond to a continued fraction expansion?)

denisrosset:
Thanks for pushing this forward!

examples/src/main/scala/spire/example/simplification.scala
erikerlandson: There is a surprising (to me) amount of Builder, SeqLike, etc, code in several of the examples.  I didn't expend much time thinking about it but it seemed a bit over-designed.

denisrosset: Do not hesitate to open an issue specifying which files should be simplified. That said, I'm not very familiar with the current examples -- I just noticed that many apply to quite abstract mathematical problems.

erikerlandson:
I'm seeing intermittent inf-looping or some other kind of hanging in the unit tests. On 2.13, the tests may run successfully or may hang partway through. `ComplexTest` or `LawTests` are possible culprits. Unsure if this is something to do with how the spire code is building or if it is caused by the scalatest/scalacheck code.

denisrosset:
Does it happen during continuous integration, or also while running the tests locally?

denisrosset:
Does it happen during continuous integration, or also while running the tests locally?

erikerlandson:
> Does it happen during continuous integration, or also while running the tests locally?
I'm seeing it with local runs too.

denisrosset:
`LawTests` is a big beast, could you divide and conquer to identify the culprit?

erikerlandson:
at the moment I'm playing with 
```
Test / logBuffered := false
Test / parallelExecution := false
```
I'm not convinced it's the tests themselves. It seems to always hang after the last test that runs in a test class.

erikerlandson:
OTOH I'm also suspicious that `Test / logBuffered := false` is actually being respected

erikerlandson:
tut is failing.  is that progress?

erikerlandson:
Looks like tut problems with 2.13 are a known issue: https://github.com/tpolecat/tut/issues/246

erikerlandson:
A summary of quirks unresolved on this PR:

- tut is broken on 2.13, so doc generation is currently disabled for 2.13 here.
- `SeqLike` and friends are defined via shim for cross-compiling. A more future-oriented solution would be to migrate signatures to `SeqOps`. This is do-able but it has impacts on how the signatures are currently defined that I would not want to introduce unilaterally. Additionally those changes might break `@specialized`.
- `TraversableOnce` is still in use. This is deprecated in favor of Iterable/IterableOnce, however the monoid definitions inherited from algebra & cats are also still using Traversable and so for now that forces the issue.
- A small number of unit tests do not compile, and a couple more seem to induce infinite looping. `.pow` and `.nroot` methods appear to be involved in all of these, so there's a good lead on where to start on that diagnosis. Currently I just have these commented out.
- the `simplification.scala` example is also currently just commented out, due to cross compile incompatabilities around the pre 2.13 `newBuilder` method.

erikerlandson:
OK, the infinite looping is a BigDecimal regression: https://github.com/scala/bug/issues/11152

larsrh:
Thanks for digging into this! Truly heroic ☺

Ichoran:
I'll fix it really soon!  (It's already fixed, but I'll get a proper test into the PR.)  Or I won't, if it's not a range thing, and you will have to set a sane precision with `MathContext` to get reasonable behaviors.

erikerlandson:
@Ichoran thank you!  If I set a `MathContext` now, will that also work on 2.13.0-M5 ?

erikerlandson:
I don't have a definitive answer for the diverging implicit expansions around `Sized[]` in the unit tests, but this seems like one possible lead:
https://github.com/retronym/scala/commit/78047b3cdce422a1e3f051c56b3fe288c88790ec#diff-7c03397456d3b987549fd039d6b639c6R960

erikerlandson:
The `Sized` compile errors are a new import conflict with `sized` method from scalacheck, renaming it seems to fix it.

```
[info] /home/eje/git/spire/tests/src/test/scala/spire/math/RealCheck.scala:111:12: <sized: error> is not a valid implicit value for org.scalacheck.Arbitrary[spire.math.ArbitrarySupport.Sized[Int,spire.math.ArbitrarySupport.Ordinal._1,spire.math.ArbitrarySupport.Ordinal._10]] because:
[info] reference to sized is ambiguous;
[info] it is imported twice in the same scope by
[info] import ArbitrarySupport._
[info] and import Gen._
[info]     forAll { (x0: Real, k: Sized[Int, _1, _10]) =>
```

erikerlandson:
picking up some additional inf-looping, which I am pretty sure has to do with Rational and its use of BigDecimal

erikerlandson:
Now not so sure that https://github.com/scala/bug/issues/11152 is the problem here, but definitely seems to have something to do with constructing `BigDecimal` instances.

erikerlandson:
I have now resolved the remaining compile errors.  There are two remaining unit tests that are inf-looping for scala 2.13. I have these turned off at run-time when running 2.13, and they will now continue to run pre 2.13.

@denisrosset, @larsrh, I am ready to propose this for review and hopefully merging. I believe the remaining loose ends are best addressed by filing dedicated issues:
- Use of deprecated Traversable[Once], here and in algebra and cats
- Tracking down inf-looping in unit tests w.r.t. 2.13
- eventual SeqLike -> SeqOps
- re-enabling tut when it works for 2.13

denisrosset:
I had a look, fine for me; if you can open issues to track the following:
- Use of deprecated Traversable[Once], here and in algebra and cats (in Spire and in cats/algebra if it is not present)
- Tracking down inf-looping in unit tests w.r.t. 2.13 (we'll probably revisit this through the 2.13 milestones)
- Re-enabling tut when it works for 2.13
I'd be happy to merge. Thanks a lot for this chunk of work!

- For the eventual SeqLike -> SeqOps, I'm not too worried. The main use cases of `Seq` are in examples, and in the vector space stuff which is not principled.

build.sbt
```diff
@@ -275,14 +274,22 @@ addCommandAlias("validate", ";validateJVM;validateJS")
...
 
 lazy val commonDeps = Seq(libraryDependencies ++= Seq(
   "org.typelevel" %%% "machinist" % machinistVersion,
-  "org.typelevel" %%% "algebra" % algebraVersion))
+  "org.typelevel" %%% "algebra" % algebraVersion,
+  "org.scala-lang.modules" %%% "scala-collection-compat" % "0.3.0"))
 
 lazy val commonSettings = Seq(
+  logBuffered := false,
+  parallelExecution := false,
```
larsrh: Do we still need this?

compat/src/main/scala-2.13/package.scala
```diff
@@ -0,0 +1,19 @@
+package spire
+package object scalacompat {
+
+import scala.collection.mutable.ArrayBuilder
+import scala.reflect.ClassTag
+
+  def arrayBuilderMake[T]()(implicit tag: ClassTag[T]): ArrayBuilder[T] =
+      ArrayBuilder.make[T]
+
+  def parallelSeq[A](s: Seq[A]): Seq[A] = {
```
larsrh: Add an issue for this too, please. Eventually I'd like to get rid of all parallel collections and replace it by `Future`-based parallelism.

build.sbt
erikerlandson: No, I'll remove that

larsrh:
This is epic! Thank you so much, @erikerlandson :tada:

erikerlandson:
@larsh, my pleasure!  Can we publish a "-M5" package build for this?

larsrh:
I'm travelling at the moment, so it'll have to wait for a few days. But I can equip @denisrosset with the publishing rights.

build.sbt
```diff
@@ -543,8 +548,9 @@ lazy val warnUnusedImport = Seq(
     CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((2, 10)) =>
         Seq()
-      case Some((2, n)) if n >= 11 =>
+      case Some((2, n)) if ((n >= 11) && (n <= 12)) =>
```
SethTisue: just a silly style thing, but you can `case Some((2, 11 | 12))`. I sometimes forget that `|` is usable other than at the top level of the match.

erikerlandson: I had no idea!

dwijnand: There's also

```scala
import sbt.librarymanagement.{ SemanticSelector, VersionNumber }
VersionNumber(scalaVersion.value).matchesSemVer(SemanticSelector("2.11.x || 2.12.x"))
```

PR #769, opened 2019-02-24
spire is on Typelevel now

larsrh:


larsrh:
>  Also: how is the release process of Spire affected by the Typelevel move? As I understand, the artifacts were already migrated, who got release privileges?

I do. We can get you the permission too.

PR #794, opened 2019-05-05
Use the `Perm` class rather than `Map[Int, Int]` for permutation instances

stewSquared:
Original rationale from gitter: https://gitter.im/typelevel/spire?at=5cc9d4011cd0b8307da5e566

A `Perm` case class already exists, as well as instances for `Map[Int, Int]` interpreting such as permutations. This PR modifies those instances to work against `Perm` rather than `Map`. This avoids potential conflicts with other interpretations of map (eg., the Map monoid from Cats), allows us to place instances in the implicit scope, and move common idioms into `Perm` helper methods. 

This PR is best read commit-by-commit.
 - Use `Perm` instead of `Map[Int, Int]`, rename `mapIntIntPermutation`
 - Make `Perm` a class extending `(Int => Int)`, move instances to companion
 - overload `compose` and `andthen` for use with other `Perm`
 - implement `inverse` and `permute` as methods and use in `Group` instance
 - Make `Perm.apply` the left action to be more consistent with [the literature](https://en.wikipedia.org/wiki/Group_action_(mathematics)#Left_group_action) and existing [semigroup instances](https://github.com/typelevel/cats/blob/b2f11931370a4525d2233c3fbfb635af66024ea0/core/src/main/scala/cats/instances/function.scala#L152) using `compose` to `combine`
 - Fix various minor issues with mapIntIntPerm method definitions


core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,62 @@
...
+import spire.syntax.group._
+import spire.util._
+
+/**
+ * Represents a permutation encoded as a map from preimages to images, including
+ * only pairs that are moved by the permutation (so the identity is Map.empty).
+ */
+case class Perm(mapping: Map[Int, Int]) extends AnyVal {
+  def apply(k: Int): Int = mapping.getOrElse(k, k)
+  def image: Set[Int] = mapping.keySet
```
stewSquared: Technically, the image is also the preimage, as assumed by line [53 below](https://github.com/typelevel/spire/pull/794/files#diff-f10faa3e5d33a83c27a54a22b3bed65eR53).

I could add an assertion that `mapping.keySet == mapping.values.toSet`

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,62 @@
...
+  def invert(k: Int) = mapping.find(_._2 == k).fold(k)(_._1)
```
stewSquared: Effectively the same as `perm.inverse.apply`

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,62 @@
...
+import spire.syntax.cfor._
+import spire.syntax.group._
+import spire.util._
+
+/**
+ * Represents a permutation encoded as a map from preimages to images, including
+ * only pairs that are moved by the permutation (so the identity is Map.empty).
+ */
+case class Perm(mapping: Map[Int, Int]) extends AnyVal {
+  def apply(k: Int): Int = mapping.getOrElse(k, k)
```
stewSquared: A permutation is a bijective function (hence the companion `invert` method).
I have half a mind to have this extend `Int => Int` and override `compose` and `andThen`.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,62 @@
...
+  def image: Set[Int] = mapping.keySet
+  def invert(k: Int) = mapping.find(_._2 == k).fold(k)(_._1)
+}
+
+object Perm {
+  implicit val PermEq: Eq[Perm] = Eq.fromUniversalEquals
```
stewSquared: This is new. Formerly the test that needed an `Eq[Map[Int, Int]]` was getting it from [`vectorOrder`](https://github.com/typelevel/spire/blob/3446d41fbbf3c0a117b4b5c86fe0ec2e7d539fd5/core/src/main/scala/spire/optional/vectorOrder.scala#L17)

stewSquared: Technically, a Map(1 -> 1, 2 -> 2, ..., n -> n) should be the same as an empty Map (they're all the identity), but the group combine method eliminates those redundant cases. I could make it a requirement of the Perm constructor, rather than normalizing here.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,62 @@
...
+
+final class PermIntAction extends Action[Int, Perm] {
+  def actr(k: Int, perm: Perm): Int = perm(k)
+  def actl(perm: Perm, k: Int): Int = perm.invert(k)
+}
+
+final class PermGroup extends Group[Perm] {
+  def empty: Perm = Perm(Map.empty[Int, Int])
+  def combine(x: Perm, y: Perm): Perm = Perm {
+    val preimages = x.image | y.image
```
stewSquared: Subtle change. Formerly concatenated, which adds every pair redundantly in the worst case scenario. (A common case, if you're combining two permutations from the same symmetric group).

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+ * only pairs that are moved by the permutation (so the identity is Map.empty).
+ */
+class Perm private(val mapping: Map[Int, Int]) extends (Int => Int) {
+  assert(mapping.forall(Function.tupled(_ != _)), "Perm contains a redundant mapping.")
+  require(mapping.values.toSet == mapping.keySet, "Image and preimage must be the same.")
+
+  override def apply(k: Int): Int = mapping.getOrElse(k, k)
+
+  override def toString: String = {
+    mapping.toSeq.sorted
```
stewSquared: Provided this `.toString` is worthwhile in the first place, we might want to use `SortedMap` rather than sorting in `.toString`, or forget about sorting altogether (doesn't really with performance anywhere; this is somewhat cosmetic).

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+import spire.algebra.partial.PartialAction
+import spire.syntax.cfor._
+import spire.syntax.group._
+import spire.util._
+
+/**
+ * Represents a permutation encoded as a map from preimages to images, including
+ * only pairs that are moved by the permutation (so the identity is Map.empty).
+ */
+class Perm private(val mapping: Map[Int, Int]) extends (Int => Int) {
```
erikerlandson: `mapping: Map[Int, Int]` is technically an implementation detail, and as such maybe should not be exposed as `val`. Perhaps `Perm` should be a trait?

erikerlandson: Is a permutation a function `Int => Int`, or a `PartialFunction`?

stewSquared: Good point. Originally, this was a case class with an exposed `map` element.
I could make this `private[Perm]` and move the `Group.combine` implementation that depends on it.

stewSquared: It's total. A value not in the Map is mapped to itself.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+  override def apply(k: Int): Int = mapping.getOrElse(k, k)
+
+  override def toString: String = {
+    mapping.toSeq.sorted
+      .map { case (k, v) => s"$k -> $v"}
+      .mkString("Perm(", ", ", ")")
+  }
+
+  def image: Set[Int] = mapping.keySet
+  def invert(k: Int): Int = mapping.find(_._2 == k).fold(k)(_._1)
```
erikerlandson: When you write the doc, I'd recommend calling out that this `invert` method is inefficient, in the sense of using linear search on each call. Contrast with `PermGroup.inverse` that generates the efficient inverse perm.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+import spire.syntax.cfor._
+import spire.syntax.group._
+import spire.util._
+
+/**
+ * Represents a permutation encoded as a map from preimages to images, including
+ * only pairs that are moved by the permutation (so the identity is Map.empty).
+ */
+class Perm private(val mapping: Map[Int, Int]) extends (Int => Int) {
+  assert(mapping.forall(Function.tupled(_ != _)), "Perm contains a redundant mapping.")
```
erikerlandson: this might now be redundant with `Perm.apply` filter below

core/src/main/scala/spire/optional/Perm.scala
erikerlandson: I see how this represents the group of permutations over all integers (or all Int).
For people using it only w.r.t. the subset that is actually permuted, it allows a composition of permutations whose domains and ranges don't align. That is probably fine, and checking it would subtract from its current elegance, but bikeshedding mathematical semantics is sort of spire's brand so... :smile: 
Maybe also just a matter of documentation.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  def image: Set[Int] = mapping.keySet
+  def invert(k: Int): Int = mapping.find(_._2 == k).fold(k)(_._1)
+}
+
+object Perm {
+  def apply(mapping: Map[Int, Int]): Perm = {
+    new Perm(mapping.filter(Function.tupled(_ != _)))
+  }
+  def apply(pairs: (Int, Int)*): Perm = apply(Map(pairs: _*))
```
erikerlandson: filter out identity mappings here?

core/src/main/scala/spire/optional/Perm.scala
stewSquared: Yeah, I'd like restricted perms too, but that would require more overhead (include identity mappings?), and depart from the previous api. For what it's worth, AlaSc and GAP both use the same semantics. Perhaps one could use a type tag like `Perm @@ S3` after checking that the image is contained by 0 until 3.

I'll make the doc a bit more explicit about image vs domain.

core/src/main/scala/spire/optional/Perm.scala
stewSquared: Yeah, we'd need a real BiMap implementation to make this efficient. I'll make sure to note the performance in the doc.

core/src/main/scala/spire/optional/Perm.scala
stewSquared: If so, they'd be redundant with the other apply method.

core/src/main/scala/spire/optional/Perm.scala
stewSquared: Actually, on second thought, that might be surprising to people expecting similar performance from `actl` and `actr`. It might be worth caching the inverse mapping privately.

core/src/main/scala/spire/optional/Perm.scala
stewSquared: Good catch; this isn't needed anymore.
Perhaps I should turn this into a property test? (The risk is that this causes `===` to fail, so I could test equality of identity mappings rather than the specific implementation detail).

core/src/main/scala/spire/optional/Perm.scala
erikerlandson: I had a similar issue one time, and IIRC I made an internal inverse, but I made it lazy, so it'd only be instantiated in the event someone called the `inverse` method

core/src/main/scala/spire/optional/Perm.scala
erikerlandson: Oh, yes, I see

core/src/main/scala/spire/optional/mapIntIntPermutation.scala
```diff
@@ -1,52 +0,0 @@
...
-    }
-  }
-  def inverse(a: Map[Int, Int]): Map[Int, Int] = a.map(_.swap).toMap
-}
-
-final class MapIntIntSeqPartialAction[A, SA <: SeqLike[A, SA]](implicit cbf: Factory[A, SA]) extends PartialAction[SA, Map[Int, Int]] {
-  import mapIntIntPermutation._
-  def partialActl(perm: Map[Int, Int], sa: SA): Opt[SA] = {
-    if (perm.isEmpty) return Opt(sa)
-    if (perm.keys.max >= sa.size) return Opt.empty[SA]
```
stewSquared: This had a bug, since negative keys would fail at runtime with `IndexOutOfBound`.

This also means we should decide whether negatives are allowed in Perms. The current Gen/Arb implementation doesn't generate them. I'm inclined to disallow them; any permutation ought to be able to permute a seq.

erikerlandson: restricting it to integers >= 0 also seems reasonable to me, fwiw

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,80 @@
...
+
+final class PermGroup extends Group[Perm] {
+  def empty: Perm = Perm(Map.empty[Int, Int])
+  def combine(x: Perm, y: Perm): Perm = Perm {
+    val preimages = x.image | y.image
+
+    preimages.foldLeft(Map.empty[Int, Int]) {
+      case (prevMap, preimage) =>
+        val image = y(x(preimage))
+        if (preimage != image) prevMap + ((preimage, image)) else prevMap
```
stewSquared: Just noticed this conditional is now unnecessary.
This whole method could probably just become `Perm(preimages.map(pre => pre -> y(x(pre))).toMap)`

core/src/main/scala/spire/optional/mapIntIntPermutation.scala
stewSquared: That's consistent with other computational group theory systems, so I'll make it a constraint in the constructor :+1:

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+    mapping.toSeq.sorted
+      .map { case (k, v) => s"$k -> $v"}
+      .mkString("Perm(", ", ", ")")
+  }
+
+  def image: Set[Int] = mapping.keySet
+  def invert(k: Int): Int = inverseMapping.getOrElse(k, k)
+  def inverse: Perm = new Perm(inverseMapping)
+
+  def compose(that: Perm): Perm = Perm((this.image | that.image).map(k => k -> this(that(k))).toMap)
```
stewSquared: This was formerly defined via `Group.combine`.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  def andThen(that: Perm): Perm = that.compose(this)
+
+  def permute[A, SA <: SeqLike[A, SA]](seq: SA)(implicit cbf: Factory[A, SA]): Opt[SA] = {
```
stewSquared: This was formerly defined via the group partial action on `SeqLike`

stewSquared:
@erikerlandson I've added docs, the non-negative constraint we discussed, and moved the definitions of `permute` and `compose` out of the typeclass instances and into methods (typeclasses now defer to these methods).

I've resolved all the comment threads that aren't just me pointing out things that have changed, so if this all looks good to you, I can clean up the commit history for a merge :smile:

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,112 @@
...
+    }
+    Opt(builder.result())
+  }
+}
+
+object Perm {
+  def apply(mapping: Map[Int, Int]): Perm = {
+    new Perm(mapping.filter(Function.tupled(_ != _)))
+  }
+  def apply(pairs: (Int, Int)*): Perm = apply(Map(pairs: _*))
```
stewSquared: My remaining qualm with this PR is that this constructor can start to become redundant.
We might rather use a constructor that takes advantage of disjoint cycle notation. eg., this perm:
```scala
Perm(
  1 -> 3,
  2 -> 4,
  3 -> 5,
  4 -> 1,
  5 -> 2
)
```
Could be more simply expressed as
```scala
Perm(1, 3, 5, 2, 4)
```

stewSquared: We could define the latter with:
```scala
  def apply(points: Int*): Perm = {
    if (points.isEmpty) Perm(Map.empty[Int, Int])
    else Perm(points.zip(points.tail :+ points.head).toMap)
  }
```
But that has the disadvantage that it can only express permutations with a single cycle.
Dennis Rosset has a nice approach in alasc that uses apply to continue composing cycles.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Represents a permutation encoded as a map from preimages to images, including
+ * only pairs that are moved by the permutation (so the identity is Map.empty).
+ *
+ * Note that although the preimage and image contain only moved points, the
+ * domain and range of a Perm is all non-negative integers.
+ */
+class Perm private(private val mapping: Map[Int, Int]) extends (Int => Int) {
+  require(mapping.values.toSet == mapping.keySet, "Image and preimage must be the same.")
```
tkroman: Would it make sense to move checks into smart constructors since actual constructor is private?

stewSquared: Ohhh, yes. That's a good idea. Then we could optimize all the internal methods to skip checks.

erikerlandson:
@stewSquared, lgtm  o/

core/src/main/scala/spire/optional/Perm.scala
stewSquared: I've gone ahead and implemented this with two required Ints and a repeated Int parameter:

```scala
def apply(n0: Int, n1: Int, ns: Int*): Perm
```

This prevents some erroneous or redundant constructions with the extra benefit of avoiding ambiguity with apply overloads (the constructor using repeated pairs: `apply(pairs: (Int, Int)*)`, and the apply method on a single int `apply (n: Int): Int`). These ultimately defer to safer constructors that enforce properties, so this should be uncontroversial.

core/src/main/scala/spire/optional/Perm.scala
```diff
@@ -0,0 +1,120 @@
...
+  implicit val PermEq: Eq[Perm] = new Eq[Perm] {
+    def eqv(p: Perm, r: Perm): Boolean = p.mapping == r.mapping
+  }
+  implicit val PermIntAction: Action[Int, Perm] = new PermIntAction
+  implicit val PermGroup: Group[Perm] = new PermGroup
+  implicit def PermSeqPartialAction[A, CC[A] <: SeqLike[A, CC[A]]](implicit cbf: Factory[A, CC[A]]): PartialAction[CC[A], Perm] = new PermSeqPartialAction[A, CC[A]]
+}
+
+final class PermIntAction extends Action[Int, Perm] {
+  def actl(perm: Perm, k: Int): Int = perm(k)
```
stewSquared: I've made `perm.apply` the left action. Formerly, it was the right action. This is more consistent with our syntax for apply, which has the function name on the left.

This also makes our `Group[Perm].combine` consistent with the `MonoidK[Endo]` instance for Function1, which uses `compose` to combine. Without this change, confusing discrepancies could arise.

stewSquared:
Alright! I've made some final changes, cleaned up the commit history, updated the PR message, and rebased against the new Scala 2.13 compat stuff.

I'm pretty happy with the result and ready for approval from maintainers.

This breaks compatibility with the old `spire.optional.mapIntIntPermutation`, so  you might want to queue it up for the next minor release.

larsrh:
Thanks a lot for the PR, @stewSquared! I'll try to get it reviewed ASAP.

core/src/main/scala/spire/optional/Perm.scala
larsrh: I say this is good enough.

stewSquared:
@larsrh Thanks! I'm glad to help and looking forward to the next release :)

PR #795, opened 2019-05-18
Remove scala-collection-compat

travisbrown:
The 1.0.0 release of scala-collection-compat is broken and we're waiting on an incompatible 2.0.0, but it's barely used here, and can be completely replaced with two type aliases and one implicit class (and the implicit class is only needed pre-2.13).

This change is fully compatible with previous releases and avoids forcing a transitive dependency on users.

r? @larsrh

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -284,8 +284,7 @@ lazy val buildSettings = Seq(
 
 lazy val commonDeps = Seq(libraryDependencies ++= Seq(
   "org.typelevel" %%% "machinist" % machinistVersion,
-  "org.typelevel" %%% "algebra" % algebraVersion,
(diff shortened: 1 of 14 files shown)


travisbrown:
Note that this change is also ~necessary if Spire wants to publish an 0.16.x release for Scala 2.13.0-RCx, since scala-collection-compat 0.3.0 is not published for 2.13.0-RC1 (and presumably won't be for RC2), and 0.3.0 to 1.0.0 is a breaking update.

larsrh:
The breakage in `scala-collection-compat` is annoying, but given that an infrastructure module is already present here, I think the loss of the dependency is acceptable.

PR #796, opened 2019-05-19
Update to 2.13.0-RC1

travisbrown:
This currently depends on a locally-published Algebra with [this change](https://github.com/typelevel/algebra/pull/228).

The ScalaTest 3.0 → 3.1 migration is even more of a nightmare here than it has been in most of the other projects where I've had to suffer through it, and everything's going to break again when we go from scalatestplus-scalacheck 1.0.0-SNAP4 to 1.0.0-SNAP5, which repackages everything yet again. I've stuck with 1.0.0-SNAP4 for now, though, to stay aligned with the ScalaTest used in the latest Cats, Discipline, etc. releases.

I've tried to clean up unused test imports in all the files I had to touch anyway, but at some point we should probably do a more thorough pass over these warnings (there are a ton of them).


travisbrown:
Okay, this is ready for review now that Algebra 2.0.0-M1 is out.

tests/src/test/scala/spire/math/NumberTest.scala
```diff
@@ -1,19 +1,16 @@
...
+import org.scalatest.check.ScalaCheckDrivenPropertyChecks
+import org.scalatest.funsuite.AnyFunSuite
 
-class NumberPropertiesTest extends PropSpec with Matchers with GeneratorDrivenPropertyChecks {
+class NumberPropertiesTest extends PropSpec with Matchers with ScalaCheckDrivenPropertyChecks {
   property("Number.apply(Long)") {
     forAll { (n: Long) => Number(n) shouldBe n }
     forAll { (n: Long) => Number(n) shouldBe SafeLong(n) }
     // we need to do (n - 1).abs to ensure we don't get a negative number
-    forAll { (n: Long) => Number((n - 1).abs) shouldBe Natural((n - 1).abs) }
```
larsrh: Why is this removed?

travisbrown: Weird, must have been a `sed` mistake or something. Good catch. I'll double-check the rest to make sure nothing else disappeared.

travisbrown:
I'll fix the Scoverage issue on 2.13.0-RC1 tomorrow.

travisbrown:
Note that in https://github.com/typelevel/spire/pull/796/commits/83429375755a6bcf4c1befe2bbf1cbe39f831cff I had to change the `BigStream` example a little for 2.13.0-RC1 because of differences between `IterableLike` and 2.13's `IterableOps` that can't easily be papered over. For what it's worth scala-collection-compat wouldn't help here, either, but luckily it's just a few lines of version-specific code. /cc @non

larsrh:
Thanks much!

PR #802, opened 2019-05-29
Replace Heyting by DeMorgan for Trilean

vquintin:
Trilean is not a valid `Heyting` as it does not satisfy the
non-contradiction law:
```
and(not(a), a) == Trilean.False
```

This commit removes the `Heyting` instance and replaces it by a
`DeMorgan` instance.

### DIFF
--- a/laws/src/main/scala/spire/laws/LogicLaws.scala
+++ b/laws/src/main/scala/spire/laws/LogicLaws.scala
@@ -2,10 +2,9 @@ package spire
 package laws
 
 import spire.algebra.{Eq, Bool}
-import spire.algebra.lattice.{DeMorgan, Heyting}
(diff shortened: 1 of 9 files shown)


laws/src/main/scala/spire/laws/LogicLaws.scala
```diff
@@ -83,6 +84,35 @@ trait LogicLaws[A] extends Laws {
       "(0 → x) = 1" -> forAllSafe { (x: A) => (A.zero imp x) === A.one }
     )
 
+  def deMorgan(implicit A: DeMorgan[A]) =
```
larsrh: Do we still need this here, given that the law definitions are also in algebra?

larsrh:
I know it's been a while, but finally we could merge this :wink:

PR #821, opened 2019-08-03
Update to discipline-scalatest-1.0.0-M1

rossabaker:
discipline-core and discipline-scalatest are no longer on the same release cycle, to decouple core's release cycle from the supported test frameworks'.  The simplest thing is to depend on discipline-scalatest and get discipline-core transitively.

Replaces #819, #820.


larsrh:
Thanks!

PR #829, opened 2019-09-04
Improve performance of root finding for big integers.

nikitaDanilenko:
The computation of the n-th root of big integers can be improved in a similar fashion as the same computation for longs. This has some performance benefits, since for large roots (say, 100) this reduces the number of necessary iterations to one-hundredth of the original number.

This optimisation in turn has the benefit of improving the speed of the fpow computation for reals, since these use big integer roots internally. For instance, computing the 1000th root of 1.5 takes 137000 iterations in the current implementation, but can be improved to only 137 iterations.

The commits handle this improvement and add a test that checks that the discrete root of big integers satisfies the desired properties.


denisrosset:
Thanks! that's a nice contribution with an accompanying test, and a direct +1 to merge!

PR #868, opened 2019-12-29
Fix badges in Readme

ybasket:
Fixes the badges for Travis, Gitter and Codecov.io as they still point to the outdated non/spire repo path


larsrh:
Thanks a lot!

PR #892, opened 2020-03-27
bump discipline-scalatest version

SethTisue:
having this upgrade merged would be helpful to me in the Scala
community build, where we can only support a single version of
discipline-scalatest (without contortions, anyway)


larsrh:
Thanks!

PR #897, opened 2020-04-08
Fix #896

### LINKED ISSUE #896: `Opt.Eq` can cause NullPointerException
The `cats.kernel.Eq` instance for `spire.util.Opt` doesn't properly handle cases where only the right hand side is empty. It will pass down a null value to the underlying `Eq[A]`, which can cause inconsistencies and/or null pointer exceptions. I have been able to produce both [in this scastie](https://scastie.scala-lang.org/Q4hMGnFySviyRU5JI2yFJQ).

Looking at the [code](https://github.com/typelevel/spire/blob/0ee38b7abc9a42fe92a63c654498305ec80be454/util/src/main/scala/spire/util/Opt.scala#L19), we can see why that is the case:
```scala
if (x.isEmpty) y.isEmpty else ev.eqv(x.ref, y.ref)
```

If `x` is non-empty then the second branch is taken. Even if `y.ref` is `null`. Then, ev.eqv is called with null in its second argument.

My suggested fix would be
```scala
if (x.isEmpty) y.isEmpty else y.nonEmpty && ev.eqv(x.ref, y.ref)
```

Props go to [Dotty explicit nulls](https://dotty.epfl.ch/docs/reference/other-new-features/explicit-nulls.html) for helping me find this issue. (Yes I was porting `Opt` to Scala 3)

lourkeur:
Address issue in the least disruptive way possible.

### DIFF
--- a/tests/src/test/scala/spire/util/OptCheck.scala
+++ b/tests/src/test/scala/spire/util/OptCheck.scala
@@ -39,6 +39,22 @@ class OptCheck extends AnyFunSuite {
     assert(eq.neqv(Opt(true) ,Opt.empty))
   }
 
+  test("Regression test for #896"){
+    {
+      import spire.std.boolean._
+      val eq = Eq[Opt[Boolean]]
+      assert(eq.neqv(Opt(false), Opt.empty))
+    }
+    {
+      import spire.std.array._
+      import spire.std.unit._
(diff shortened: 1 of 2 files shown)


denisrosset:
Thanks @lourkeur !

PR #909, opened 2020-05-18
Reincorporated Machinist macros into Spire.

denisrosset:


larsrh:
I don't understand this change. Why not remove the macros altogether like in Cats (typelevel/cats#2925)?

denisrosset:
Because Cats allocates like mad for the `Ops` traits, at least under the YJ profiler

larsrh:
This is going to be a blocker for Dotty, though. Not sure what to do about it.

denisrosset:
Ok... I'm running this without the YJP agent, on the IntelliJ debugger, and it does **not** allocate.

Profiler agents seem to block escape analysis: http://psy-lob-saw.blogspot.com/2014/12/the-escape-of-arraylistiterator.html

On Dotty of course, the story is more dire: we won't have specialization. But I guess performance sensitive parts of code bases would be then kept on Scala 2; but we'll still need the syntax glue on Dotty... without macros then?

larsrh:
Incorporated into #910 (feel free to reopen if this is not correct)

PR #915, opened 2020-06-19
Bugfix for Trig[Complex[_]]

gregor-i:
`Trig[Complex[A]].asin` uses `sin` instead of `asin`
`Trig[Complex[A]].acos` uses `cos` instead of `acos`
`Trig[Complex[A]].atan` uses `sin` instead of `atan`


denisrosset:
Nice job spotting this!

PR #910, opened 2020-05-18
Bump Scala and Scala.js versions, bump algebra, cross publish SJS 0.6, drop 2.11 support

denisrosset:

PR #920, opened 2020-07-13
Skip coverage and update scala.js

cquiroz:
This updates seem to make travis happy and includes these updates:

* Update scala.js to 1.1.1 and makes it the default
* Skip coverage check
* Skip publication of the jvm side if scala.js is set to 0.6.x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ TAGS
 .history
 .*.swp
 .idea
+project/metals.sbt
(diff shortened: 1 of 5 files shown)


cquiroz:
I made this as a PR to #910 but it could be converted to a PR to master

tpolecat:
Sanity-checked this with @larsrh and looks fine.

PR #924, opened 2020-08-06
Switch to LazyList

NthPortal:
Add APIs that use `LazyList` instead of `Stream`, and
deprecate `Stream` APIs.

Change examples to use `LazyList` instead of `Stream`.

Does not address `BigStream` implementation in
examples.

### DIFF
--- a/core/src/main/scala/spire/math/Real.scala
+++ b/core/src/main/scala/spire/math/Real.scala
@@ -2,6 +2,8 @@ package spire
 package math
 
 import scala.math.{ScalaNumber, ScalaNumericConversions}
+import scala.collection.compat.immutable.LazyList
+import scala.collection.compat.immutable.LazyList.#::
 
 import spire.algebra.{Field, Trig, TruncatedDivisionCRing}
 import spire.syntax.nroot._
@@ -474,7 +476,19 @@ object Real extends RealInstances {
 
   lazy val sqrt1By2 = Real.two.reciprocal.sqrt
 
+  def accumulate(total: SafeLong, xs: LazyList[SafeLong], cs: LazyList[Rational]): SafeLong = {
+    (xs, cs) match {
+      case (_, Seq()) => total
+      case (Seq(), _) => sys.error("nooooo")
+      case (x #:: xs, c #:: cs) =>
+        val t = roundUp(c * Rational(x))
+        if (t == 0) total else accumulate(total + t, xs, cs)
+    }
+  }
+
+  @deprecated("prefer LazyList instead", "0.17.0")
   def accumulate(total: SafeLong, xs: Stream[SafeLong], cs: Stream[Rational]): SafeLong = {
+    import scala.#::
     (xs, cs) match {
       case (_, Stream.Empty) => total
       case (Stream.Empty, _) => sys.error("nooooo")
@@ -484,7 +498,7 @@ object Real extends RealInstances {
     }
   }
 
-  private[spire] def powerSeries(ps: Stream[Rational], terms: Int => Int, x: Real): Real = {
+  private[spire] def powerSeries(ps: LazyList[Rational], terms: Int => Int, x: Real): Real = {
     Real({p =>
       val t = terms(p)
       val l2t = 2 * sizeInBase(SafeLong(t) + 1, 2) + 6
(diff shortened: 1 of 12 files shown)


core/src/main/scala/spire/math/poly/SpecialPolynomials.scala
```diff
@@ -2,14 +2,25 @@ package spire
...
+import scala.collection.compat.immutable.LazyList
 
 import spire.algebra.{Eq, Field, Ring}
 import spire.math.Polynomial
 import spire.syntax.field._
 
 object SpecialPolynomials {
 
+  // Horner scheme polynomial generator lazy list
+  def hornerSchemeLazy[C: Ring: Eq: ClassTag](zero: Polynomial[C], one: Polynomial[C],
```
NthPortal, line 14: I'm not fond of appending `Lazy` to all the names, but sadly there are no `LazyList` parameters to overload on. open to suggestions for a better naming scheme (could make a separate object with a different name for them? I guess? idk if that's any better)

core/src/main/scala/spire/random/Dist.scala
```diff
@@ -138,7 +143,7 @@ trait Dist[@sp A] extends Any { self =>
     rawHistogram(n).map { case (k, v) => (k, 1.0 * v / n) }
 
   def rawHistogram(n: Int)(implicit gen: Generator): Map[A, Int] =
-    toStream(gen).take(n).foldLeft(Map.empty[A, Int]) { case (h, a) =>
+    toLazyList(gen).take(n).foldLeft(Map.empty[A, Int]) { case (h, a) =>
```
NthPortal, line 146: I'm not sure if there's value in constructing a `LazyList` over this or if `toIterator` would be more appropriate, since it's not being cached (unless I've missed something)

core/src/main/scala/spire/math/Real.scala
```diff
@@ -474,7 +476,19 @@ object Real extends RealInstances {
...
+      case (Seq(), _) => sys.error("nooooo")
+      case (x #:: xs, c #:: cs) =>
+        val t = roundUp(c * Rational(x))
+        if (t == 0) total else accumulate(total + t, xs, cs)
+    }
+  }
+
+  @deprecated("prefer LazyList instead", "0.17.0")
   def accumulate(total: SafeLong, xs: Stream[SafeLong], cs: Stream[Rational]): SafeLong = {
+    import scala.#::
```
larsrh, line 491: Why is this import here?

core/src/main/scala/spire/math/Real.scala
```diff
@@ -2,6 +2,8 @@ package spire
 package math
 
 import scala.math.{ScalaNumber, ScalaNumericConversions}
+import scala.collection.compat.immutable.LazyList
```
larsrh, line 5: I'm assuming this can be shortened to `scala.collection.immutable.LazyList` when we drop pre-2.13 support?

core/src/main/scala/spire/math/poly/SpecialPolynomials.scala
larsrh: To be honest, I don't think we necessarily need to keep two versions of the same methods here. I don't think there are that many users of `SpecialPolynomials`.

core/src/main/scala/spire/random/Dist.scala
larsrh: No clue honestly, but using `toLazyList` seems safer.

core/src/main/scala/spire/math/Real.scala
NthPortal: so... there is an `object #::` in the `scala` package object containing `unapply` for each of `LazyList` and `Stream` that forward to their respective extractors. unfortunately, in 2.12-, it's just an alias for `Stream`'s extractor, since `LazyList` didn't exist. Because of this, I had to import `LazyList.#::` at the top of the file to make *its* extractor work. That import ends up shadowing `scala.#::`, making `Stream`'s extractor no longer work, so I had to add this import. I wanted `LazyList` to be the "natural" one, but we could also remove this import and move the `LazyList.#::` import to a localised scope where it doesn't affect the `accumulate` overload for `Stream`

core/src/main/scala/spire/math/Real.scala
NthPortal: it can be dropped entirely, in fact; `LazyList` (like `List` and `Vector`) has type and object aliases in the default scope

core/src/main/scala/spire/math/poly/SpecialPolynomials.scala
NthPortal: I wasn't sure how important binary compatibility between versions was. I'm happy to remove the `Stream` versions

larsrh: spire is currently in 0.x, so we don't make any guarantees :wink: (that is to say, the deprecation cycle for the other methods is fine, but in this particular object I don't see a problem).

NthPortal:
squashed

larsrh:
I'll just go ahead and merge this. Thanks for your contribution!

Issue #921, opened 2020-07-23
-Rational.one triggers deprecation warning about auto-application in 2.13.3


lunaryorn:
Scala 2.13.3 deprecates auto-application of methods with empty parameter lists, see https://github.com/scala/scala/pull/8833.  Spire defines  the unary `-` of `Rational` with an [empty parameter list](https://github.com/typelevel/spire/blob/88efc919765fb50370d63cfe4de8ebc871b4857b/core/src/main/scala/spire/math/Rational.scala#L49).  Consequently negating a rational (i.e. `-Rational.one`) triggers the new deprecation warning, e.g. (with `import spire.math.{Rational => R})`:

```
Error:(49, 25) Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method unary_-,
or remove the empty argument list from its definition (Java-defined methods are exempt).
In Scala 3, an unapplied method like this will be eta-expanded into a function.
      SnapToGrid(R.one, -R.one, Closed(R.zero))
```

I'm not sure whether this can be considered a bug in Scala 2.13.3, but it looks as if just removing the empty parameter list from `spire.match.Rational.unary_-` (i.e. defining it without any argument list) would fix this issue on the Spire side.  Other numeric types in Spire (e.g. [`UByte`](https://github.com/typelevel/spire/blob/88efc919765fb50370d63cfe4de8ebc871b4857b/core/src/main/scala/spire/math/UByte.scala#L54)) already define the unary - without parameter lists.

larsrh:
Yeah, good point. This will be an error in Dotty. Would you mind sending a PR?

lunaryorn:
To remove these parenthesis?  Can do next week, please ping me on eg Thursday should I forget 🙂

larsrh:
I mean, there's many other occurrences of `unary_-()`, so help would be appreciated :wink:

larsrh:
Should be fixed now.

PR #932, opened 2020-08-24
Remove unary ops parentheses

leoniv:
I faced with same problem as https://github.com/typelevel/spire/issues/921. I executed `grep -e 'unary_.()' -r core/` maybe it's enough or not i don't know


larsrh:
Thanks a bunch!

PR #974, opened 2021-01-25
Fewer warnings

cquiroz:
Spire has quite a bunch of warnings (several hundreds) which would be great IMHO to remove before going into dotty.
This PR does remove a large amount of them.

I have divided this PR into two commits. The first are trivial changes while the second contain some less trivial that maybe wort checking

Some are still there though most of them could be removed it it were to become 2.13 only, WDYT?


core/src/main/scala/spire/compat.scala
```diff
@@ -7,7 +7,7 @@ import spire.math.{ScalaIntegralWrapper, ScalaNumericWrapper, ScalaPartialOrderi
 private[spire] trait CompatPriority1 {
   implicit def numeric[A: Ring: ConvertableFrom: Signed: Order]: scala.math.Numeric[A] =
     new ScalaNumericWrapper[A] {
-      val partialOrder = PartialOrder[A]
+      // val partialOrder = PartialOrder[A]
```
larsrh, line 10: Should probably be removed

core/src/main/scala/spire/macros/fpf/Fuser.scala
```diff
@@ -13,10 +13,10 @@ private[spire] trait Fuser[C <: Context, A] {
   private def Epsilon: Tree = q"2.220446049250313E-16"
   private def PositiveInfinity: Tree = q"java.lang.Double.POSITIVE_INFINITY"
   private def NegativeInfinity: Tree = q"java.lang.Double.NEGATIVE_INFINITY"
-  private def isNaN(a: TermName): Tree = q"java.lang.Double.isNaN($a)"
-  private def isInfinite(a: TermName): Tree = q"java.lang.Double.isInfinite($a)"
+  // private def isNaN(a: TermName): Tree = q"java.lang.Double.isNaN($a)"
+  // private def isInfinite(a: TermName): Tree = q"java.lang.Double.isInfinite($a)"
```
larsrh, line 17: same here

core/src/main/scala/spire/macros/fpf/Fuser.scala
```diff
@@ -13,10 +13,10 @@ private[spire] trait Fuser[C <: Context, A] {
...
   private def max(a: Tree, b: Tree): Tree = q"java.lang.Math.max($a, $b)"
-  private def min(a: Tree, b: Tree): Tree = q"java.lang.Math.min($a, $b)"
+  // private def min(a: Tree, b: Tree): Tree = q"java.lang.Math.min($a, $b)"
```
larsrh, line 19: same here

core/src/main/scala/spire/math/Complex.scala
```diff
@@ -31,7 +31,7 @@ object Complex extends ComplexInstances {
     bigDecimalToComplex(BigDecimal(n))
 
   implicit def bigDecimalToComplex(n: BigDecimal): Complex[BigDecimal] = {
-    implicit val mc = n.mc
+    // implicit val mc = n.mc
```
larsrh, line 34: same here

core/src/main/scala/spire/math/Number.scala
```diff
@@ -492,7 +492,7 @@ private[math] case class DecimalNumber(n: BigDecimal) extends Number { lhs =>
   def r_/(lhs: Number): Number = Number(lhs.toBigDecimal / n)
   def r_tquot(lhs: Number): Number = Number(lhs.toBigDecimal quot n)
   def r_tmod(lhs: Number): Number = Number(lhs.toBigDecimal % n)
-  private def tuplize(t: (BigDecimal, BigDecimal)) = (DecimalNumber(t._1), DecimalNumber(t._2))
+  // private def tuplize(t: (BigDecimal, BigDecimal)) = (DecimalNumber(t._1), DecimalNumber(t._2))
```
larsrh, line 495: same here

core/src/main/scala/spire/math/Number.scala
```diff
@@ -575,7 +575,7 @@ private[math] case class RationalNumber(n: Rational) extends Number { lhs =>
   def r_tquot(lhs: Number): Number = Number(lhs.toRational tquot n)
   def r_tmod(lhs: Number): Number = Number(lhs.toRational tmod n)
 
-  private def tuplize(t: (Rational, Rational)) = (RationalNumber(t._1), RationalNumber(t._2))
+  // private def tuplize(t: (Rational, Rational)) = (RationalNumber(t._1), RationalNumber(t._2))
```
larsrh, line 578: same here

core/src/main/scala/spire/math/Polynomial.scala
```diff
@@ -541,8 +541,10 @@ trait PolynomialOverField[@sp(Double) C] extends PolynomialOverRing[C]
         val ys = y match {
           case yd: poly.PolyDense[C] => poly.PolySparse.dense2sparse(yd)
           case ys1: poly.PolySparse[C] => ys1
+          case _ => sys.error("no")
```
larsrh, line 544: I'd rather have an `unchecked` annotation at the scrutinee (same for other places where you added `sys.error`)

core/src/main/scala/spire/math/Sorting.scala
```diff
@@ -36,7 +36,7 @@ object InsertionSort extends Sort {
     * @param end the index of the last element, exclusive, to be sorted
     * @tparam A a member of the type class `Order`
     */
-  final def sort[@sp A](data:Array[A], start:Int, end:Int)(implicit o:Order[A], ct:ClassTag[A]): Unit = {
+  final def sort[@sp A: ClassTag](data:Array[A], start:Int, end:Int)(implicit o:Order[A]): Unit = {
```
larsrh, line 39: could be combined into `: ClassTag : Order`

core/src/main/scala/spire/math/Sorting.scala
```diff
@@ -197,7 +197,7 @@ object QuickSort {
     * @tparam A a member of the type class Order
     * @return the next pivot value
     */
-  final def partition[@sp A](data:Array[A], start:Int, end:Int, pivotIndex:Int)(implicit o:Order[A], ct:ClassTag[A]): Int = {
+  final def partition[@sp A: ClassTag](data:Array[A], start:Int, end:Int, pivotIndex:Int)(implicit o:Order[A]): Int = {
```
larsrh, line 200: same here

core/src/main/scala/spire/math/package.scala
```diff
@@ -142,8 +142,8 @@ package object math {
       return power(BigDecimal(1), part, whole.toBigInt)
     }
 
-    var precision = k.mc.getPrecision + 3
-    var leeway = 1000
+    // var precision = k.mc.getPrecision + 3
```
larsrh, line 145: delete commented out code

core/src/main/scala/spire/math/poly/RootIsolator.scala
```diff
@@ -48,8 +48,8 @@ object RootIsolator {
    * goes into greater detail.
    */
   private final def VAS(poly: Polynomial[BigInt]): Vector[Interval[Rational]] = {
-    val x = Polynomial.x[BigInt]
-    val one = Polynomial.one[BigInt]
+    // val x = Polynomial.x[BigInt]
```
larsrh, line 51: same here

core/src/main/scala/spire/math/prime/Siever.scala
```diff
@@ -56,7 +56,7 @@ case class Siever(chunkSize: Int, cutoff: SafeLong) {
     if (n == 3) return SafeLong(2)
 
     var i = 3
-    var k = n - 1
+    // var k = n - 1
```
larsrh, line 59: same here

core/src/main/scala/spire/random/Dist.scala
```diff
@@ -294,24 +294,24 @@ object Dist extends DistInstances9 {
...
+  implicit val ubyte: Dist[UByte] = new DistFromGen[UByte](g => UByte(g.nextInt()))
+  implicit val ushort: Dist[UShort] = new DistFromGen[UShort](g => UShort(g.nextInt()))
+  implicit val uint: Dist[UInt] = new DistFromGen[UInt](g => UInt(g.nextInt()))
+  implicit val ulong: Dist[ULong] = new DistFromGen[ULong](g => ULong(g.nextLong()))
 
   implicit def complex[A: Fractional: Trig: IsReal: Dist]: Dist[Complex[A]] =
     Dist(Complex(_: A, _: A))
 
-  implicit def interval[A](implicit na: Dist[A], order: Order[A], r: AdditiveMonoid[A]): Dist[Interval[A]] =
+  implicit def interval[A: AdditiveMonoid](implicit na: Dist[A], order: Order[A]): Dist[Interval[A]] =
```
larsrh, line 314: `: AdditiveMonoid : Dist : Order`

core/src/main/scala/spire/random/rng/Marsaglia32a6.scala
```diff
@@ -80,7 +80,7 @@ class Marsaglia32a6(_x: Int, _y: Int, _z: Int, _w: Int, _v: Int, _d: Int) extend
 
 object Marsaglia32a6 extends GeneratorCompanion[Marsaglia32a6, Array[Int]] {
   def fromBytes(bytes: Array[Byte]): Marsaglia32a6 = {
-    val bs = if (bytes.length < 24) Arrays.copyOf(bytes, 24) else bytes
+    // val bs = if (bytes.length < 24) Arrays.copyOf(bytes, 24) else bytes
```
larsrh, line 83: delete commented code

core/src/main/scala/spire/std/map.scala
```diff
@@ -64,7 +64,7 @@ class MapVectorSpace[K, V](override implicit val scalar: Field[V]) extends MapCR
   override def times(x: Map[K, V], y: Map[K, V]): Map[K, V] = {
     var xx = x
     var yy = y
-    var f = scalar.times _
+    // var f = scalar.times _
```
larsrh, line 67: same here

larsrh (changes requested):
I only have nitpicks, otherwise looking good!

larsrh:
> Some are still there though most of them could be removed it it were to become 2.13 only, WDYT?

I'd rather do that in a separate PR that removes 2.12 support.

core/src/main/scala/spire/math/Rational.scala
```diff
@@ -690,7 +690,8 @@ object Rational extends RationalInstances {
 
     override def equals(that: Any): Boolean = that match {
       case that: LongRational => this.n == that.n && this.d == that.d
-      case that => super.equals(that)
+      case rat: Rational => rat ==that
+      case that => this == that
```
larsrh: This change looks a bit fishy; possibly prone to endless recursion.

cquiroz: Yes I had to revert this. I'm keeping as it was and may add a @nowarn once 2.12 is removed

core/src/main/scala/spire/math/Polynomial.scala
cquiroz: the `sys.error("no")` comes because it was used on a few places before this PR. but I could also use the annotation

core/src/main/scala/spire/math/Sorting.scala
cquiroz: The warning here is because the variable is not in use, It maybe better to unify the style and use always all as implicit vals or in the type param

core/src/main/scala/spire/math/Polynomial.scala
larsrh: That's Erik's humour then, I guess :) I'd still prefer the annotation.

core/src/main/scala/spire/math/Sorting.scala
larsrh: If the variable is not in use, couldn't we get rid of the context bound too?

core/src/main/scala/spire/math/Polynomial.scala
cquiroz: Actually the original one was `sys.error("nooooooo")` :smile:

core/src/main/scala/spire/math/Sorting.scala
cquiroz: The warning complains that the variable `ct` is unused but the implicit ClassTag is used so we need the context bound.

I tried not to touch too much the implicits because in some cases they are not really used in the function but to constraints the callers. The most obvious one is when a `NoImplicit` is required

larsrh: That's weird, but in that case, I guess we can keep it as is.

cquiroz:
I added the suggested fixes. Do you think I should rather squash to have a single commit with the changes?

larsrh:
Awesome, thanks! I'll just squash-merge.

larsrh (approved):

PR #978, opened 2021-01-30
Drop support for scala 2.12

cquiroz:
This PR drops supports for 2.12 and simplifies the build somewhat removing the compatibility layers used to make it compile in 2.12 and 2.13

I have to admit feeling bad ripping the compat layer which probably took some effort to create

An extra benefit is that the build now has no warnings


benchmark/src/main/scala/spire/benchmark/ComplexAddBenchmarks.scala
```diff
@@ -1,14 +1,14 @@
...
+//
+// import scala.util.Random
+// import Random._
+// import org.openjdk.jmh.annotations._
+// import spire.algebra._
+// import spire.math._
+// import spire.implicits._
+// import spire.benchmark.Arrays.init
 
 /*
```
larsrh, line 13: Have you tried compiling this benchmark?

benchmark/src/main/scala/spire/benchmark/ZigguratBenchmarks.scala
```diff
@@ -15,7 +15,6 @@
 package spire
 package benchmark
 
-import spire.implicits._
 /*
```
larsrh, line 18: Same here

examples/src/main/scala-2.13/spire/example/BigStreamCompat.scala
```diff
@@ -9,5 +9,5 @@ trait BigStreamCompat[A] extends IterableOps[A, BigStream, BigStream[A]]
 
 trait BigStreamCompanionCompat extends IterableFactory[BigStream] {
   def from[A](source: IterableOnce[A]): BigStream[A] =
-    source.foldLeft(BigStream.empty[A])((t, a) => new BigCons(a, t))
+    source.iterator.foldLeft(BigStream.empty[A])((t, a) => new BigCons(a, t))
```
larsrh, line 12: Examples directory should be just `scala` now

benchmark/src/main/scala/spire/benchmark/ComplexAddBenchmarks.scala
cquiroz: Indeed, there are several commented out benchmarks. Maybe I can do that in a separate PR and if I can't just delete them?

larsrh: Sounds good :+1:

cquiroz:
This is now rebased and fixed a few details commented on the PR

larsrh:
Let's go :+1:

PR #1027, opened 2021-05-21
Type ascriptions

cquiroz:
This PR is towards scala 3 support. Scala 3 require `implicit` to have an explicit type ascription and this is included in this PR

### DIFF
--- a/core/src/main/scala/spire/math/Number.scala
+++ b/core/src/main/scala/spire/math/Number.scala
@@ -609,7 +609,14 @@ private[math] case class RationalNumber(n: Rational) extends Number { lhs =>
 }
 
 trait NumberInstances {
-  implicit final val NumberAlgebra = new NumberAlgebra
+  implicit final val NumberAlgebra: Field.WithDefaultGCD[Number]
+    with CRing[Number]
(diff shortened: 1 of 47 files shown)


core/src/main/scala/spire/math/Number.scala
```diff
@@ -609,7 +609,14 @@ private[math] case class RationalNumber(n: Rational) extends Number { lhs =>
 }
 
 trait NumberInstances {
-  implicit final val NumberAlgebra = new NumberAlgebra
+  implicit final val NumberAlgebra: Field.WithDefaultGCD[Number]
```
milessabin, line 612: Why isn't this `implicit final val NumberAlgebra: NumberAlgebra = new NumberAlgebra`?

cquiroz: I'm following the style of cats like:
https://github.com/typelevel/cats/blob/main/kernel/src/main/scala/cats/kernel/instances/BigIntInstances.scala#L5

milessabin: Fair enough ... I don't know why it's done that way there.

PR #1045, opened 2021-07-14
Rename master to main

danicheg:
This is a minor thing but should be common for all `typelevel` projects, I guess.


armanbilge:
@rossabaker can you help with this? I don't have sufficient privilege to do the branch rename (bit of a relief tbh :)

Issue #851, opened 2019-11-20
Remove the Checked macros


denisrosset:
Few code paths in Spire use the `Checked` machinery. For the Scala 2 -> 3 transition we'd be better off reducing our use of macros.

armanbilge:
`Checked` survived the transition! https://github.com/typelevel/spire/pull/1067

PR #1057, opened 2021-08-19
Port macros/util/platform to build in Scala 3

cquiroz:
This is a first PR to support Scala 3. The port is a very large task considering the amount of macros and features supported by spire but some progress has been done

I've managed to compile `core` in scala 3 but that still needs some work and I want to make sure all tests are passing. In the interim I can send this PR that cross builds three base packages to scala3

Note that I don't plan to port all macros, just the minimum, thus in this PR I only include a port of `Checked`. machinist is not ported but core can be built without it using directly typeclasses

I added a temporal git hub action to build these three modules with scala 3 but will removed once the full port is done


armanbilge:
@cquiroz thanks for your tremendous efforts on this! What are your thoughts on https://github.com/typelevel/spire/issues/851, which suggests that the `Checked` macro should be dropped?

cquiroz:
Spire has a bunch of macros and my efforts on the scala 3 work has been predicated in dropping most of them. `checked` is the exception that I think should be kept as it really needs to take a look at the structure of the code inside and macros is the way to go

Another premise for the scala 3 port is that for the most part we don't change the semantics of spire

cquiroz:
BTW I found a small bug on my checked macros implementation, we should wait before merging this one

cquiroz:
I've hit a few edge cases with `Checked` in particular with `tryOrReturn` which is itself kind of on the edge. I'm thinking to delete it and rewrite them as
```scala
try {
   Checked.checked(blah)
} catch { (_: ArithmeticOverflowException) => return something}
```
At least in the meanwhile this will do

armanbilge:
@cquiroz I'm still not convinced we shouldn't just drop it as @denisrosset suggested. I appreciate all your efforts and don't mean to trivialize your work, but could you help justify why it's worth all this effort?

> `checked` is the exception that I think should be kept

cquiroz:
Would you suggest to drop it without a replacement? just letting overflows be? That would be fairly easy but would change the semantics quite a bit?

Or would you rewrite the code in each place `Checked` is used to verify overflows?

armanbilge:
[When I search "Checked"](https://github.com/typelevel/spire/search?q=Checked) I find it used in two places:
- `SafeLong`: either drop this in Scala 3 or maybe we can find an easy way to re-implement.
- `laws/shadows`: [when I search "shadows"](https://github.com/typelevel/spire/search?q=shadows) in only seems to be used by itself. So can we just not port this package to Scala 3?

Apologies if I am missing something important.

cquiroz:
`Rational` has several uses of `Checked.tryOrElse` and `Checked.tryOrReturn`

cquiroz:
Some examples
https://github.com/typelevel/spire/blob/main/core/src/main/scala/spire/math/Rational.scala#L97-L110

This is also quite involved:
https://github.com/typelevel/spire/blob/main/core/src/main/scala/spire/math/Algebraic.scala#L1352-L1432

armanbilge:
I see, thank you for the examples. I'm sorry that I missed those. Now I understand better why `Checked` is so important. Still, I wonder what @denisrosset had in mind.

cquiroz:
I've dropped most of the other macros (I kept a few that could be easily replaced by inlines) as it seems they were mostly done for performance reasons (and could be restored later on) but I can't easily see how `Checked` could be removed without semantic changes. 

I hope this is enough of a justification.

armanbilge:
> but I can't easily see how Checked could be removed without semantic changes.

I agree, I do see that it is used in very important code now, and we should definitely avoid semantic changes 100%.

> I hope this is enough of a justification.

Absolutely, I'm sorry for asking an annoying question. I was confused since this contradicted the previous maintainers' ideas.

cquiroz:
Superseded by #1067

PR #1067, opened 2021-09-16
Dotty

cquiroz:
## Scala 3 port
This PR adds support for scala3 while still cross compiling to scala 2.13.
It is a fairly complex PR due to the structure of spire and its extensive use of techniques such as macros and typeclass based syntax that need to be changed for scala 3

The aim is to have a cross compiled version with no or minimal changes to the API and keep a reasonable performance
The mark of success will be passing all (or most tests) while the API remains mostly unchanged. 
While this is totally usable we may need to do some further refinements after this is merged

## General changes:
* do/while constructs were converted to while
* Implicits need type ascription
* Some number conversions need to be explicit
* Remove or rewrite the macros

## API changes
In general we want clients not needing to change their code at all, however
in some cases it was unavoidable:

* the method `Dist.given` clashes with the `given` keyword, it was renamed to `given_`
* the class `DistIterator` has a `next` member that clashes with a `next` method. The field has renamed to `_next`
* `UnboundSyntax` seems to be bogus and it is not used anywhere in `spire`. It was marked as deprecated in scala 2 and removed in scala 3
* It is not in this PR yet but we should change all the empty paren methods to no-parens

## Syntax changes
`spire` contains a fairly large set of syntaxis to allow a variety of operations. These use macros and needed to be rewritten. 

As part of the process I changed them to be mostly extension methods.
However this changes the priorities as seen here:
https://github.com/lampepfl/dotty/issues/8740

I discovered this late in the process as it only has effects in some cases where all syntax is
imported and there are multiple implementations of the same operator.

I propose to test the approach in this PR but could be reverted if necessary

## Macros
Spire includes a fair amount of macros many of them fairly complex.
As a first approximation I applied this policy about macros:

**Only port macros that are strictly necessary to keep the semantics not the ones related to performance**:

Under that criteria the action for each macro is as follow:
* **Checked**: This is likely the most complicated macro. It can take an arbitrary expression and check if any of
the int/long operations would cause an overflow. It is hard to replace this with a non macro as it is supposed
to operate with any arbitrary expression.
In a first run `Checked.tryOrReturn` was removed as it can't be expressed as a `quasiquote` and in general it
is hard to reason about when used
Uses of `tryOrReturn` have been rewritten with the general form of

```scala
try {
  Checked.checked(a)
} catch {
  case _: ArithmeticException => return b
}
```

* **PackMacro**: This are fairly simple thus were converted using scala3 `inline`. Note that part of
`Pack` is now shared with the macros being scala version specific

[...]

PR #1078, opened 2021-10-18
fix publish workflow

i10416:
Add workflow to publish latest document microsite.

https://typelevel.org/spire/ does not show latest document.


.github/workflows/publish_microsite.yml
```diff
@@ -0,0 +1,33 @@
+name: Publish Doc
+
+on:
+  pull_request:
```
armanbilge, line 4: Is this publishing the docs for every PR??

armanbilge:
Thanks btw! We have some big changes in progress in https://github.com/typelevel/spire/pull/1067, and there's also #1069 to fix.

i10416:
@armanbilge I will take a look at #1069 and try to send another PR.

.github/workflows/publish_microsite.yml
```diff
@@ -0,0 +1,32 @@
...
+          fetch-depth: 0
+      - name: Setup Jekyll
+        run: |
+          gem install --install-dir $HOME/vendor/bundle jekyll
+          echo "GEM_HOME=$HOME/vendor/bundle" >> $GITHUB_ENV
+          echo "GEM_PATH=$HOME/vendor/bundle" >> $GITHUB_ENV
+          echo "$HOME/vendor/bundle/bin" >> $GITHUB_PATH
+      - name: Build docs
+        run: |
+          sbt docs/makeMicrosite
```
armanbilge, line 26: Where does `sbt` come from? I think you usually need to install it with an action.

Instead of as a hand-written workflow, Do you think we can include this in the sbt build with sbt-gh-actions? Then most of the setup and compiling is done automatically. See e.g. https://github.com/typelevel/fs2/blob/main/build.sbt#L349

Thanks!

i10416: >Where does sbt come from?

ubuntu-latest has sbt installed.

i10416: @armanbilge 

I choose a hand-written workflow because I have one concern.  Is it possible for sbt-gh-actions to define multiple workflow? 

It is faster to publish docs when we use multiple workflows because workflow runs in parallel but steps do not run in parallel.

i10416: @armanbilge 

> Do you think we can include this in the sbt build

Answer is "yes, we can" but do we really need that?

I think hand-written workflow is better when it comes to managing spire document because

- current publish task has little to do with Scala(only `mdoc` task), it is done by `jekyll` and `actions-gh-pages`. Wrapping them in sbt task does not seem straightforward.
- I took a look `sbt-gh-actions` and `sbt-gh-actions` focus mainly on making it easier to test, build and publish Scala artifacts **over cross java/scala versions and cross target platforms(jvm/native/js)** (though it is possible to add publish docs jobs).
  - In addition, if we include publish task  in jobs, we have to introduce another action to detect file changes, which can be done by github workflow's path filter.
- some libraries in typelevel repo like `cats-effect` uses hand-written workflow for publishing docs. https://github.com/typelevel/cats-effect/blob/docs/.github/workflows/microsite.yml
- separating document settings from build settings will keep build settings simple(and makes it easier to migrate to another document tools in the future.)

You say "setup and compiling is done automatically" but these features make little difference for publishing docs. We do not have to setup Java or Scala for publishing docs. We need setup `jekyll`, envs and paths even if we use `sbt-gh-actions`. 

If we publish docs for each spire versions over each Scala versions and each platforms it is worth managing it in build.sbt, but current microsite is independent of Scala versions or platforms, just displaying information of the latest version, so the publish doc workflow is simple enough and will not be likely to swell further.

i10416: Ah, you mean you want to programatically garantee docs are successfully built using 'if ....' or 'needs build' ? If so, it would be better to include  the doc job in build.sbt.

Sorry, I assumed that docs are supposed to be successfully built everytime PR is merged to main (or version tag is set) because current workflow contains build docs step and you can be sure docs are vaild.

> So we should actually be careful to only publish the site when we release a new, stable version

Ok, workflow becomes simpler and there's no need to check doc file changes as publishing documents happens only when stable release, not when doc files changes.

armanbilge: Yes, just for my own clarification the docs should be _built_ as part of every PR, to avoid breaking something. And as you point out, they already are. But of course, they should not be _published_ from the PR :)

> Ok, workflow becomes simpler and there's no need to check doc file changes as publishing documents happens only when stable release, not when doc files changes.

The downside to this is that docs contributions/updates are not published until the next release ... but, it prevents the API docs from getting out of sync with the published artifact, which could be very confusing. I'm open to ideas to improve this.

One way would be to publish the API docs under a version directory (e.g., 0.17, 0.18, SNAPSHOT). So this way, even though updating the site would also update the API docs, it would publish under SNAPSHOT instead of overwriting the API docs for the current.

For clarification, I'm distinguishing between the site/main docs (which should ideally be updated as frequently as possible) and the API docs aka scaladocs, which should match the published artifact.

Thoughts? Thanks again for your help with this, much appreciated!

PR #1084, opened 2021-10-19
Update scalafmt to 3.0.6

armanbilge:
Scala 3 sources needed a dialect override because they are using Scala 3 only syntax that is not compatible with the `Scala213Source3` dialect.


benchmark/src/main/scala/spire/benchmark/ZigguratBenchmarks.scala
```diff
@@ -1,15 +1,8 @@
...
- * *  /_____/ /_/      /_/   /_/  |_|  /____/     All rights reserved.    **
- * *                                                                      **
- * *      Redistribution and use permitted under the MIT license.         **
- * *                                                                      **
+ * **********************************************************************\ * Project ** * ______ ______ __ ______ ____
+ * ** * / ____/ / __ / / / / __ / / __/ (c) 2011-2014 ** * / /__ / /_/ / / / / /_/ / / /_ ** * /___ / / ____/ / / / __ /
+ * / __/ Erik Osheim, Tom Switzer ** * ____/ / / / / / / / | | / /__ ** * /_____/ /_/ /_/ /_/ |_| /____/ All rights
+ * reserved. ** * ** * Redistribution and use permitted under the MIT license. ** * **
  * \***********************************************************************
  */
```
armanbilge, line 7: Oh no 🤦

armanbilge: follow up in #1086

cquiroz: I think you can indicate a block doesn't need formatting
https://scalameta.org/scalafmt/docs/configuration.html#-format-off

armanbilge: Yes, but I've never seen anybody use that for their headers. Maybe this situation is different, but I'd prefer a better fix if possible.

core/src/main/scala-3/spire/syntax/Syntax.scala
```diff
@@ -269,50 +271,50 @@ trait LiteralsSyntax {
       sb.append(lits.next())
       while (vars.hasNext)
         sb.append(vars.next())
-        sb.append(lits.next())
+      sb.append(lits.next())
```
armanbilge, line 274: Oh crap.

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
```diff
@@ -94,54 +96,54 @@ def fastForRangeMacro(r: Expr[Range], body: Expr[Int => Unit])(using quotes: Quo
...
     }
+  }
 
   def strideUpTo(fromExpr: Expr[Int], untilExpr: Expr[Int], stride: Expr[Int]): Expr[Unit] = '{
     var index = $fromExpr
-    val end   = $untilExpr
+    val end = $untilExpr
     while (index <= end)
-      ${ Expr.betaReduce(body) }(index)
-      index += $stride
```
armanbilge, line 110: hmm??

cquiroz: I would expect `scalafmt` not to change the semantics of the code

armanbilge: Yeah seems like a bug. 😠

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
```diff
@@ -3,24 +3,26 @@ package spire.syntax.macros
...
 
-import spire.syntax.fastFor.{RangeLike, RangeElem}
+import spire.syntax.fastFor.{RangeElem, RangeLike}
 
 inline def fastForInline[R](init: R, test: R => Boolean, next: R => R, body: R => Unit): Unit =
   var index = init
   while (test(index))
     body(index)
-    index = next(index)
+  index = next(index)
```
cquiroz: Hmm, this changes the code

core/src/main/scala-3/spire/syntax/Syntax.scala
```diff
@@ -71,7 +71,9 @@ trait AdditiveSemigroupSyntax:
     new LiteralDoubleAdditiveSemigroupOps(lhs)
 
 trait AdditiveMonoidSyntax extends AdditiveSemigroupSyntax:
-  implicit def additiveMonoidOps[A](a: A)(implicit ev: AdditiveMonoid[A]): AdditiveMonoidOps[A] = new AdditiveMonoidOps(a)
+  implicit def additiveMonoidOps[A](a: A)(implicit ev: AdditiveMonoid[A]): AdditiveMonoidOps[A] = new AdditiveMonoidOps(
```
cquiroz, line 74: Maybe a line break?

PR #1095, opened 2021-10-23
Improve docs

### LINKED ISSUE #1080: v0.18.0 migration guide
documentation for changes mentioned here https://github.com/typelevel/spire/pull/1067#issue-998607764

i10416:
close #1080 
-  ~~move  md files to docs/src/main/mdoc~~  
- ~~add scala 3 migration guide~~ CHANGES.md and release notes are duplicated and we move to github release notes.
- [x] update outdated information
- [x] fix some dead links
- [x] inprove docs especially for new users


i10416:
@armanbilge  ping

I've finish updating documents.  Sorry for taking time since I opened this PR.

Could you review PR when you have time?

build.sbt
```diff
@@ -330,30 +332,11 @@ lazy val docSettings = Seq(
...
-    file("AUTHORS.md") -> ExtraMdFileConfig(
-      "authors.md",
-      "home",
-      Map("title" -> "Authors", "section" -> "Home", "position" -> "5")
-    ),
-    file("CHANGES.md") -> ExtraMdFileConfig(
-      "changes.md",
-      "home",
-      Map("title" -> "Changes", "section" -> "Home", "position" -> "2")
-    ),
```
armanbilge, line 342: I think it's helpful to continue to keep these at the top of the repo. For example cats is like this:
https://github.com/typelevel/cats

docs/src/main/mdoc/migration.md
```diff
@@ -0,0 +1,49 @@
+---
+layout: home
+title:  "scala 3 migration"
+section: "Home"
+position: 7
+---
+
+## Scala 3 Migraion Guide
```
armanbilge, line 8: Thanks for writing this up! I'm not sure this is a Scala 3 migration guide, since we still support Scala 2.13 as well. I think it's more accurately 0.17 -> 0.18 migration guide, what do you think?

Also I think the notes here are based on the PR and a little out of date. The latest is in https://github.com/typelevel/spire/releases/tag/v0.18.0-M1.

armanbilge (commented):
No worries, thank you so much for all your help around here! Took a quick pass through.

docs/src/main/mdoc/migration.md
i10416: Yes, you are right. This is a bit misleading. 

> Also I think the notes here are based on the PR and a little out of date. The latest is in https://github.com/typelevel/spire/releases/tag/v0.18.0-M1.

Thanks, I will update according to the release!

I will re-write what you pointed out soon!

build.sbt
i10416: Ah, I moved them when I tried to do this hack https://github.com/typelevel/spire/pull/1095#issuecomment-950253326

I need move them back.

i10416: Btw, I think `DESIGN.md` an `FRIENDS.md` will be likely to contain code block(e.g. for explanation or example). Do you think they should be moved back too?

i10416: (Probably you mentioned line 333 to 342 only. Just to be sure.)

docs/src/main/mdoc/migration.md
```diff
@@ -0,0 +1,52 @@
...
+
+If you find any erroneous behavior due to Scala 3 compatibility, please feel free to open issue or send PR!
+
+Spire tries keeping compatibility as much as possible, but some APIs change.
+If your code uses following APIs, you need to modify them.
+
+## API changes
+ 
+### General
+- because the method `Dist.given` clashes with the `given` keyword, it was renamed to `given_`
```
armanbilge, line 20: This is also out-of-date.

docs/src/main/mdoc/migration.md
```diff
@@ -0,0 +1,52 @@
...
+- the class `DistIterator` has a `next` member that clashes with a `next` method. The field has renamed to `_next`
+- `UnboundSyntax` seems to be bogus and it is not used anywhere in `spire`. It was marked as deprecated in scala 2 and removed in scala 3
+- parentheses: these are now omitted for 0-arity (empty-parens) methods with no side-effects.
+### Macro
+- `Checked.tryOrReturn` was removed.
+- Use of `Machinist` was removed.
+- `PackMacro` is available both Scala 2 and 3.
+- `cFor` was deprecated and renamed to `fastFor`. Available both Scala 2 (as `cFor`) and 3(as `fastFor`). Performance-sensitive code should continue using `cfor` on Scala 2.
```
armanbilge, line 28: There are some important explanations for this, it would be strange if the release notes are more detailed than the migration guide 😆

docs/src/main/mdoc/migration.md
```diff
@@ -0,0 +1,52 @@
+---
+layout: home
+title:  "migration guide"
+section: "Home"
+position: 7
+---
+
+## v0.18.0 Migraion Guide
```
armanbilge, line 8: Do you think it's helpful to have this migration guide on a separate page? What if we put it on the Changes page?
http://typelevel.org/spire/extra_md/changes.html

docs/src/main/mdoc/migration.md
```diff
@@ -0,0 +1,52 @@
...
+- Remove or rewrite the macros
+
+### Performance
+
+As of Nov 2021, Scala 3 support can cause performance degration compared to Scala 2 mainly because of lack of `specialization`. Some quick benchmarking shows no important difference for code not using `specialization`. On the other hand, code relying on this feature takes a severe performance hit. This is mentioned in [v0.18.0-M1 release](https://github.com/typelevel/spire/releases/tag/v0.18.0-M1) and [#1067](https://github.com/typelevel/spire/pull/1067#issue-998607764) too.
+
+The complete benchmark suite run for 2.13 vs 3 is available here: [#1067 (comment)](https://github.com/typelevel/spire/pull/1067#issuecomment-939369626).
+
+
+Latest information is available at [release](https://github.com/typelevel/spire/releases) page.
```
armanbilge, line 52: I don't want to trivialize your work, but since users probably want the latest information ... should we put this link at the to? 😅

build.sbt
armanbilge: Hmm, good question ... you are right `DESIGN.md` could have code blocks, but currently they do not I think.

Would it be okay with you, if we put it back?
1. Cats also has `DESIGN.md` in its top directory: https://github.com/typelevel/cats
2. I'm conservative about making changes here, since I am not an original author of the library, sorry :(

docs/src/main/mdoc/migration.md
i10416: At first, I thought it would be helpful because there are some breaking change between 0.17.0 and 0.18.0 and users can easily find them (In general, users take it for granted that there are breaking API changes between major versions like 0.x.y and 1.0.0, but may not between minors.), but after your comment and thinking twice, the migration guide in separate page is not good idea because we need to maintain CHANGES.md, migration.md and release notes at github that are almost the same. Thank you for pointing this out.

build.sbt
i10416: No worries!

docs/src/main/mdoc/migration.md
armanbilge: > maintain CHANGES.md, migration.md and release notes at github that are almost the same.

Hmm, you're right, that seems like a lot of duplication. Actually, why _aren't_ they the same? Is there something missing from the release notes, maybe we should add that there instead? Since many users may get their information from there without checking the website I think.

I think `CHANGES.md` was important before GitHub had "releases" feature. But now we that we use releases, maybe the `CHANGES.md` should just give a link to the release page for the release notes?

> In general, users take it for granted that there are breaking API changes between major versions like 0.x.y and 1.0.0, but may not between minors.

Regarding this, I'm not sure it's true ... if you look at the changelog for spire, historically there have often (always) been breaking changes between 0.x releases. Furthermore, this is common for many Scala projects using (early) semantic versioning, such as http4s, circe, etc. In fact, this is even encoded inside `sbt`, so if you try to use two different minor versions of a library in the 0.x stage it will stop you!

i10416: > I think CHANGES.md was important before GitHub had "releases" feature. But now we that we use releases, maybe the CHANGES.md should just give a link to the release page for the release notes?

I agree. That would be better. 

Which option do you like?

1. delete duplicated migration stuffs from CHANGES.md and replace them with link to release note
2. keep migration stuffs for now and add link to release note

i10416: in addition, If we move to github release note, I think it would be helpful to mention it at the beginning of CHANGES.md like "since v0.18.0, we track changes in \[github releases\]\(link/to/release/notes\)"

i10416: > why aren't they the same? Is there something missing from the release notes,

AFAIK, there's nothing missing and my migration guide may be a bit redundant😓

armanbilge: Yes 💯

I think the only change we need to make to `CHANGES.md` is to put a link to github releases exactly like that. That way, we avoid duplicating/separating information about changes that belongs together.

Then, I know you can't PR the release notes, but if there's important information you think is missing from there please let me know or propose some edits, I really appreciate having another pair of eyes on it :)

Thanks again for all your efforts!

armanbilge (approved):
Thank you!

PR #1111, opened 2021-11-17
ensure fastForInline does not make closures

bishabosha:
make the parameters of fastForInline `inline`, which will beta reduce them when applied.
Currently examining produced bytecode, closures are created for `test`, `next` and `body`


armanbilge:
Brilliant, thank you! I remember asking about this in https://github.com/typelevel/spire/pull/1067#discussion_r711654605 but seems I forgot to create an issue to track it 😅 

We're not fussed about bincompat at the moment, but for my own education: what are the bincompat implications of a change like this?

If you feel so motivated, feel free to run the [`FastForBenchmark`](https://github.com/typelevel/spire/blob/main/benchmark/src/main/scala/spire/benchmark/FastForBenchmark.scala) :)

armanbilge:
Hmm, this seems to have broken the semantics somehow 🤔 

I only looked quickly, but it's also possible you managed to restore the `cfor` semantics for this test.
https://github.com/typelevel/spire/blob/a667a9984fb25d75c7c4e6e7a641e84ba8b29a44/tests/shared/src/test/scala/spire/syntax/FastForSuite.scala#L102-L113

If that's the case, then maybe we _can_ implement `cfor` on Scala 3 with identical semantics to Scala 3 `cfor`, which would be excellent :)

cquiroz:
This seems great, I wonder why the tests broke

tests/shared/src/test/scala/spire/syntax/FastForSuite.scala
```diff
@@ -60,7 +60,7 @@ class FastForSuite extends munit.FunSuite {
         b += x
       }
     }
-    assertEquals(v, 111)
+    assertEquals(v, 334) // 4 * `test`, 3 * `next`, 3 * `body`
```
armanbilge, line 63: If we have to change these tests, then that means that `fastFor` semantics still don't match `cforSemantics` 😕

bishabosha:
I have updated the tests to reflect the fact that any of the arguments to `fastFor` / `fastForRange` can capture the loop variable.

This does mean that the semantics of the scala 2 versions does not match, so the test will probably fail again.

tests/shared/src/test/scala/spire/syntax/FastForSuite.scala
bishabosha: I see, so it appears the cfor macro only beta reduces the function argument if it is a lambda, and not with block statements before

bishabosha: I think the semantics could match if we don't implement fastFor with inline, but turn it to macro and check that the argument is a lambda.

armanbilge: Would you have the bandwidth to work on that? Appreciate all your efforts so far :)

If we can make `fastFor` behave like `cfor` again, we can do away with `fastFor` altogether which would be excellent.

armanbilge: (The purpose of the name change was to highlight the changed semantics.)

bishabosha: I added the change I described and reverted the side-effect tests to their original state

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
```diff
@@ -20,11 +20,34 @@ import collection.immutable.NumericRange
...
+def fastForImpl[R: Type](init: Expr[R], test: Expr[R => Boolean], next: Expr[R => R], body: Expr[R => Unit])(using
+  Quotes
+): Expr[Unit] =
+
+  def fastCode =
+    for
+      testOpt <- Util.tryOpt(test)
+      nextOpt <- Util.tryOpt(next)
+      bodyOpt <- Util.tryOpt(body)
+    yield '{
```
armanbilge, line 32: Forgive my ignorance, but is there a reason we need `tryOpt` to succeed for all three of these together? Would it be possible to have a "hybrid" code which uses fast code for some of these and slow code for the others?

armanbilge (commented):
🎉 hooray, all tests passing! Thank you so much 😁 

I have a question below.

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
bishabosha: this was just my first try :) I agree it would be nice to have the ability to only have one deoptimised argument - I'll see what can be done

bishabosha: I have implemented the ability to optimise function arguments independently of each other, and included it on fastForRange, allowing it to match the behaviour of cforRange

bishabosha:
@armanbilge I have optimised the implementation for both `fastFor` and `fastForRange` so that side effects will not be inlined in either. I am still unsure about the status of having `fastForRange` for `Long` as well, as that is not implemented in Scala 2

armanbilge:
Wow! I see you were able to fix the semantics so that they match `cfor` again. That's fantastic, thank you!

I think this means we can rename `fastFor` back to `cfor`, share the tests across 2/3, and remove any reference to `fastFor` (we only introduced it due to the semantic changes). Doesn't have to be in this PR (I can do this refactoring in a follow-up). Then maybe we can release 0.18.0 final 😁 

Right now the Scala 2 `fastFor` reference implementation is failing the CI. To land your PR, maybe let's make the `fastFor` test suite Scala 3-only.

I'll take a closer look later today. Thanks again for your efforts here :)

bishabosha:
@armanbilge I have made the FastForSuite scala 3 only :)

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
```diff
@@ -15,157 +15,117 @@
...
+end RangeForImpl
 
-  def strideUpTo(fromExpr: Expr[Int], untilExpr: Expr[Int], stride: Expr[Int]): Expr[Unit] = '{
-    var index = $fromExpr
-    val end = $untilExpr
-    while index <= end do
-      ${ Expr.betaReduce(body) }(index)
-      index += $stride
-  }
+private def letFunc[A, B, C](using Quotes)(name: String, rhs: Expr[A => B])(in: Expr[A => B] => Expr[C]): Expr[C] =
```
armanbilge, line 117: Would you mind writing a comment(s) to explain a little of the macro magic here? For curious users and future contributors/maintainers alike :)

armanbilge (approved):
LGTM and not so hard to wrap your head around, although `letFunc` went a bit over my head 😅 Regarding the `Long` implementation, I think it we should definitely include it 👍 

I hope @cquiroz can take a look as well :)

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
```diff
@@ -15,157 +15,117 @@
...
-          '{ val b = $body; $r.foreach(b) }
-      }
-    case _ =>
-      report.warning(s"defaulting to foreach, can not optimise range expression", r)
-      '{ val b = $body; $r.foreach(b) }
+  def let[A, B](name: String, f: Expr[A])(in: Expr[A] => Expr[B]): Expr[B] =
+    ValDef.let(Symbol.spliceOwner, name, f.asTerm)(ref => in(ref.unsafeAsExpr[A]).asTerm).unsafeAsExpr[B]
 
-end fastForRangeMacro
+  if isFunctionLiteral(rhs) then in(Expr.betaReduce(rhs))
```
cquiroz, line 130: Nice, in the old scala2 code there was a differentiation for literals as well that I didn't manage to port to scala 3

cquiroz (approved):
Looks great, thanks a lot for this fix

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
```diff
@@ -15,157 +15,117 @@
 
 package spire.syntax.macros
 
-import quoted._
-import collection.immutable.NumericRange
+import scala.quoted.*
```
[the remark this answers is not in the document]
bishabosha: It's still not needed, I thought it just made it more clear where `quoted` comes from

core/src/main/scala-3/spire/syntax/macros/cforMacros.scala
bishabosha: I have added some doc comments, and also found an extractor I can use to simplify function literal detection

bishabosha:
@armanbilge @cquiroz I have pushed another commit to simplify a bit and document

armanbilge (commented):
Thanks, looks great! I hope that as you find time, you may continue to contribute—I think spire has an interesting journey ahead as we figure out how to work in a language without specialization :)

PR #1115, opened 2021-11-28
Update to algebra 2.7.0

armanbilge:
This PR re-integrates spire with algebra 2.7.0. The following typeclasses from spire were ported into algebra:
- `GCDRing`
- `EuclideanRing`
- `Signed` / `SignedAdditiveCMonoid`/ `SignedAdditiveAbGroup`
- `TruncatedDivision` / `TruncatedDivisionCRing`
- `DivisionRing`
- `Field`

Now the spire definitions are removed and instead alias the algebra definitions.

Most of the churn here is due to the changes to `Signed`. Previously, in spire, `Signed <: Order`. However, when `Signed` was ported to algebra it no longer extends `Order`. Instead it has a member `def order: Order`. This technique is used to avoid ambiguous implicits (e.g. how `FunctorFilter` doesn't extend `Functor` in cats).

As a result of this change, an additional `Order` constraint had to be propagated across spire where it used to be provided by `Signed`.


build.sbt
```diff
@@ -20,7 +20,7 @@ lazy val scalaCheckVersion = "1.15.4"
 lazy val munit = "0.7.29"
 lazy val munitDiscipline = "1.0.9"
 
-lazy val algebraVersion = "2.2.3"
+lazy val algebraVersion = "2.7.0"
```
cquiroz, line 23: Should this be rename to `catsVersion`?

core/src/main/scala-3/spire/syntax/Ops.scala
```diff
@@ -83,7 +83,7 @@ final class SignedOps[A](a: A)(using s: Signed[A]):
   def isSignNonNegative: Boolean = s.isSignNonNegative(a)
 
 final class TruncatedDivisionOps[A](lhs: A)(using ev: TruncatedDivision[A]):
-  def toBigIntOpt: Opt[BigInt] = ev.toBigIntOpt(lhs)
+  // def toBigIntOpt: Opt[BigInt] = ev.toBigIntOpt(lhs) // TODO port to algebra?
```
cquiroz, line 86: This seems to be an API change, why?

core/src/main/scala/spire/math/Complex.scala
```diff
@@ -117,22 +117,22 @@ final case class Complex[@sp(Float, Double) T](real: T, imag: T)
    *
    * `sgn(z) = z / abs(z) = abs(z) / z`
    */
-  def complexSignum(implicit f: Field[T], n: NRoot[T], s: Signed[T]): Complex[T] =
+  def complexSignum(implicit f: Field[T], n: NRoot[T], o: Order[T], s: Signed[T]): Complex[T] =
```
cquiroz, line 120: Why is `Order` required now?

core/src/main/scala-3/spire/syntax/Ops.scala
cquiroz: Could it be implemented in `spire`?

cquiroz (approved):
Thanks a lot for this. I was expecting the change to be less extensive. Is it so because there are duplicates between algebra and spire?
Why is `Order` now required in many more places?

build.sbt
armanbilge: Yeah, `catsVersion` = `algebraVersion` now. But since we don't explicitly depend on cats, I think this is fine for now.

core/src/main/scala-3/spire/syntax/Ops.scala
armanbilge: For some reason it wasn't included in https://github.com/typelevel/algebra/pull/247.

I think it would be good if we could continue to implement it in spire, but I wasn't sure how unless we introduce a new typeclass? Any ideas?

core/src/main/scala/spire/math/Complex.scala
armanbilge: See https://github.com/typelevel/spire/pull/1115#issue-790147766 about `Order`.

armanbilge:
Because these changes are so extensive, I wonder if we should release a second milestone after this PR merges. I think we should wait to do an RC/final release until https://github.com/typelevel/spire/pull/1111 is resolved.

PR #1119, opened 2021-12-03
use blackhole to improve benchmark

i10416:
accumulating result into a local variable suffers from software
pipelining as mentioned in Jmh example about looping. See https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_34_SafeLooping.java

In short, this benchmark does not produce proper result.

```java
    @Benchmark
    public int measureWrong_2() {
        int acc = 0;
        for (int x : xs) {
            acc += work(x);
        }
        return acc;
    }
```

You should use Blackhole to avoid compiler optimization. Blackhole
forces runtime to compute every `work()` call in full.

```java
    @Benchmark
    public void measureRight_1(Blackhole bh) {
        for (int x : xs) {
            bh.consume(work(x));
        }
    }
```

### DIFF
--- a/benchmark/src/main/scala/spire/benchmark/AddBenchmarks.scala
+++ b/benchmark/src/main/scala/spire/benchmark/AddBenchmarks.scala
@@ -30,124 +30,122 @@ import org.openjdk.jmh.infra.Blackhole
 @State(Scope.Thread)
 class AddBenchmarks {
 
-
-  def addGeneric[@sp(Int, Long, Float, Double) A: Ring](data: Array[A],bh:Blackhole):Unit = {
+  def addGeneric[@sp(Int, Long, Float, Double) A: Ring](data: Array[A], bh: Blackhole): Unit = {
     var total = Ring[A].zero
     var i = 0
     val len = data.length
     while (i < len) {
-      bh.consume({total = Ring[A].plus(total, data(i))})
+      bh.consume { total = Ring[A].plus(total, data(i)) }
       i += 1
     }
   }
 
-  def addFastComplex(data: Array[Long],bh:Blackhole): Unit = {
+  def addFastComplex(data: Array[Long], bh: Blackhole): Unit = {
     var total = FastComplex(0.0f, 0.0f)
     var i = 0
     val len = data.length
-    while (i < len) { 
-      bh.consume({total = FastComplex.add(total, data(i))})
-      i+=1
+    while (i < len) {
+      bh.consume { total = FastComplex.add(total, data(i)) }
+      i += 1
     }
   }
 
   @Benchmark
-  def addIntDirect(state: IntState,bh:Blackhole): Unit = {
+  def addIntDirect(state: IntState, bh: Blackhole): Unit = {
     val data = state.values
     var total = 0
     var i = 0
     val len = data.length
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks! I need to think about this. Considering this example:
```scala
while (i < len) { total = Ring[A].plus(total, data(i)); i += 1 }
```

I agree, that if we want to measure _only_ `Ring[A].plus(total, data(i))` then we need to use blackhole, because otherwise there will be pipelining or other hotspot optimizations.

But, if actually what we want to measure is the performance of operating _on an entire Array_, then actually I think this benchmark is doing the right thing. Because actually we _want_ to see if the JVM is able to optimize this or not.

What do you think?

i10416:
Thank you for reviewing :D
I do not know why these add benchmarks use array (perhaps with the intent to measure agaist different scale of numbers? ) but even if this benchmark tried to check peformance of operating on an entire Array, I still think Blackhole makes sense.

If this benchmark tried to check whether JVM can optimize entire array operation, we need to test it on different JVMs and care about contribution rate of factors, don't we? (And I doubt it is worth it in terms of maintenance. )
 
Otherwise, I think (micro) benchmark should be independent of JVM. It would be better to distinguish how fast spire is and how well spire can be optimized by a specific JVM.

armanbilge:
Of course! Wonderful to have a regular contributor :)

> I do not know why these add benchmarks use array

I think because arrays are a very important use-case (e.g. see the polynomial algorithms in spire, also data science and machine learning algorithms, etc.). Especially when you care about performance: if you are only working with a few numbers then performance is less important. But if you are working with arrays of 100s or 1000s or 10000s of numbers, then even a small difference per-operation can make a big difference.

So it's important that our benchmarks are realistic for arrays: pipelining is an important optimization, and maybe some changes to the implementation would cause the JVM to disable this optimization for some reason (in fact, I don't even know if it _is_ pipelining right now!). But if use a blackhole to disable optimizations in the benchmarks, we won't be able to correctly measure this real-world performance.

> we need to test it on different JVMs

I agree. But IIRC effectively all modern JVMs are based on HotSpot (except Graal), so I expect the specific JVM shouldn't make much difference (besides the version number).

> Otherwise, I think (micro) benchmark should be independent of JVM.

This is impossible :) even if the blackhole prevents pipelining of consecutive instructions, the JVM will still optimize the instructions inside the blackhole.

FWIW it may still be correct to use a blackhole here, but I don't think we should be trying to disable array optimizations because I think that is what we should be trying to benchmark.

i10416:
> I don't even know if it is pipelining right

Oh😯 I don't too. I wonder how much spire takes advantage of JVM optimizations.

> This is impossible :) even if the blackhole prevents pipelining of consecutive instructions, the JVM will still optimize the instructions inside the blackhole.

Yeah, I know. I meant "as much as possible"😑

> I think that is what we should be trying to benchmark.

Ok! I withdraw PR. Thanks anyway for explanation!

armanbilge:
Btw, if you completely want to disable hotspot optimizations I think you can add the JVM flag `-Djava.compiler=NONE`. But since nobody runs JVM like that in real world, IDK how to interpret the result of that!

PR #1122, opened 2021-12-04
migrate caliper to jmh

i10416:


benchmark/src/main/scala/spire/benchmark/PowBenchmark.scala
```diff
@@ -15,79 +15,92 @@
...
     }
     t
   }
 
-  def timeDoublePowForDouble(reps:Int) = run(reps) {
+  @Benchmark
+  def timeDoublePowForDouble(): Double = {
     var t = 0.0
-    longs.foreach { n =>
+    doubles.foreach { n =>
```
armanbilge, line 101: Good catch!

armanbilge (approved):
Thanks, this was a big help!

PR #1137, opened 2022-01-07
Rename `fastFor` back to `cfor`

### LINKED ISSUE #1136: Change `fastFor` back to `cfor`
Now that in https://github.com/typelevel/spire/pull/1111 the semantics are matching across Scala 2/3, we no longer need `fastFor` :) since we only included it in milestones, I would prefer to remove it rather than do a deprecation phase prior to 0.18.0 final. Any thoughts?

armanbilge:
Closes #1136. Happy to put this behind us :)

Are we okay with no deprecation cycle on `fastFor`?


armanbilge:
Also, FTR, I do think `fastFor` had intrinsic value due to its simpler (less surprising?) semantics. But I think that is a separate discussion.

cquiroz (approved):
Excellent, lots of deleted code :smile:

PR #1173, opened 2022-05-16
Don't assume NRoot[Real] is perfect

### LINKED ISSUE #1106: `RealScalaCheckSuite.x.pow(k).nroot(k) = x` failed
https://github.com/typelevel/spire/runs/4087917370?check_suite_focus=true#step:8:1039

```
==> X spire.math.RealScalaCheckSuite.x.pow(k).nroot(k) = x  0.049s munit.FailException: /home/runner/work/spire/spire/tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala:124
123:    }
124:  }
125:

Failing seed: IhNYEXMVtFv8DyJdCDjhw3OngO0cs8m_RcJEQuIvjaG=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "IhNYEXMVtFv8DyJdCDjhw3OngO0cs8m_RcJEQuIvjaG="

Falsified after 9 passed tests.
> ARG_0: -2759766405657757/1532495540865888858358347027150309183618739122183602176
> ARG_1: Sized(6)
    at munit.Assertions.fail(Assertions.scala:283)
    at munit.Assertions.fail$(Assertions.scala:15)
    at munit.FunSuite.fail(FunSuite.scala:11)
    at munit.ScalaCheckSuite.propToTry$$anonfun$1(ScalaCheckSuite.scala:104)
    at scala.util.Try$.apply(Try.scala:210)
    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:104)
    at munit.ScalaCheckSuite.$init$$$anonfun$2$$anonfun$1$$anonfun$1(ScalaCheckSuite.scala:46)
    at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:439)
    at munit.Suite$$anon$1.execute(Suite.scala:26)
    at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:393)
    at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:302)
    at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:276)
    at scala.concurrent.impl.Promise$DefaultPromise.transform(Promise.scala:128)
    at munit.internal.FutureCompat$ExtensionFuture.transformCompat(FutureCompat.scala:16)
    at munit.ScalaCheckSuite.$init$$$anonfun$3$$anonfun$2(ScalaCheckSuite.scala:48)
    at munit.GenericTest.withBodyMap$$anonfun$1(GenericTest.scala:33)
    at munit.MUnitRunner.$anonfun$8(MUnitRunner.scala:296)
```

kschwarz1116:
Fixes #1106.

Open to other suggestions here!


tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala
```diff
@@ -112,25 +114,28 @@ class RealScalaCheckSuite extends munit.ScalaCheckSuite {
...
   property("x.pow(3) = x * x * x") {
     forAll { (x: Real) =>
-      x.pow(2) == x * x
+      x.pow(3) == x * x * x
     }
   }
 
   property("x.pow(k).nroot(k) = x") {
+    implicit val realArb: Arbitrary[Real] = Arbitrary(realFromLongs)
     forAll { (x0: Real, k: Sized[Int, _1, _10]) =>
```
armanbilge, line 123: I'm not really sure if this is any nicer, but it is a bit more explicit :)

```suggestion
    forAll(realFromLongs, arbitrary[Sized[Int, _1, _10]]) { (x0, k) =>
```

tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala
```diff
@@ -112,25 +114,28 @@ class RealScalaCheckSuite extends munit.ScalaCheckSuite {
 
   property("x.pow(3) = x * x * x") {
     forAll { (x: Real) =>
-      x.pow(2) == x * x
+      x.pow(3) == x * x * x
```
armanbilge, line 117: Woops, thank you! :)

tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala
```diff
@@ -196,28 +201,6 @@ class RealScalaCheckSuite extends munit.ScalaCheckSuite {
...
-  //     }
-  //   }
-  // }
-
-  // sample1("sample1 id")(x => x)
-  // sample1("sample1 negate")(x => -x)
-  // sample1("sample1 +")(x => x + x)
-  // sample1("sample1 *")(x => x * x)
-  // sample1("sample1 sqrt")(_.sqrt())
-  // sample1("sample1 pow(2)")(_.pow(2))
```
armanbilge, line 219: I wonder what this was. It's been commented for nearly a decade now it seems.

tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala
armanbilge: Ah, this will fix the compile errors I reckon.

kschwarz1116: Yeah, this worked in 3 but not in 2 😮 -- implicit resolution must have changed in 3.
I split these tests into a separate suite so that it's more clear.

armanbilge: I think if we explicitly pass the generator, then we can keep them in the same file.

kschwarz1116: Happy to go that route if that's standard here - I felt kind of bad writing `arbitrary[Sized[Int, _1, _10]]` 3 times though 🤷‍♂️

tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala
kschwarz1116: Your guess is as good as mine. It seemed like a unused test that I couldn't really decipher.

tests/shared/src/test/scala/spire/math/RealScalaCheckSuite.scala
armanbilge: I'm not entirely sure what `Sized` is, but you may be able to use `Gen.chooseNum` or something to achieve the same result. I think it's worth it to keep the tests together.

kschwarz1116: I'm struggling to find a solution that uses `realFromLongs` in the derivation of `Arbitrary[NonZero[Real]]` for [this case](https://github.com/typelevel/spire/pull/1173/files#diff-f8231a49bbd9e332a3e1bd93a9fdce1b9903cd499b2bf7c6c4fd0b7cda38d107R46) that is compatible between scala 2 and 3 but keeps everything in one suite. It seems like where I'd previously have used an `implicit def` in scala 2, [I should be using a `given` in scala 3](https://docs.scala-lang.org/scala3/reference/contextual/relationship-implicits.html#given-instances). I suppose two suites in one file would be doable here too, but that's not far off from what's already in the PR.

I pulled in that test here because it seemed like it was susceptible to the same problem in the issue. What do you think the path forward should be?

armanbilge: I'm not sure I understand what the Scala 2/3 issue is here, since `implicit def` should also work on Scala 3. Are you saying that `Arbitrary[NonZero[Real]]` is not using the `realFromLongs` at the top of that file?

kschwarz1116: Ok so I read the compiler output again and it's not a compat issue, I'm just not sure how to provide implicit evidence properly. Sorry about the confusion.

Here's the current `NonZero` instance:
```scala
 implicit def nonZeroSpireImplicit[A: Signed: AdditiveGroup: Arbitrary]: Arbitrary[NonZero[A]] =
    Arbitrary(arbitrary[A].filter(_.signum != 0).map(NonZero(_)))
```

I want the instance of `Arbitrary[A]` used here to be `realFromLongs`, so I updated it to:
```scala
implicit def nonZeroSpireImplicit[A: Signed: AdditiveGroup](implicit arbA: Arbitrary[A]): Arbitrary[NonZero[A]] =
    Arbitrary(arbA.arbitrary.filter(_.signum != 0).map(NonZero(_)))
```
I was hoping to pass in `realFromLongs` explicitly, but in doing so I can't find evidence for `Signed[Real]`.

armanbilge: Aha, gotcha, no worries :) IIRC before you changed it the signature was like this:
```scala
implicit def nonZeroSpireImplicit[A: Signed: AdditiveGroup: Arbitrary]: Arbitrary[NonZero[A]]
```

You can then explicitly pass in `realFromLongs` something like this:
```scala
nonZeroSpireImplicit[Real](implicitly, implicitly, Arbitrary(realFromLongs))
```

armanbilge (approved):
Looks great, thanks for iterating on this!!

PR #1182, opened 2022-06-17
Reduce min tests for native in various scalacheck suites

armanbilge:
I think this will shave like 10+ minutes off the Native build, which is pretty crazy.


armanbilge:
Yup, this seems to have reduced test from 25 minutes to 8 minutes. Native is still the slow poke in CI.

PR #1247, opened 2023-05-23
Drop implicit conversions for compat with future Scala 3 versions

smarter:
Replace `type PairedSet[A] = (Set[A], Set[A])` by a case class. The old definition relied on implicit conversions from and to `Set|A]` to support a `map` operation, but Scala 3 already defines `Tuple#map`. So far, this did not break anything because `Tuple#map` takes a polymorphic function type, but I'm working on type inference improvements that would let the user pass a regular closure and have the type block inferred, which means the implicit conversion would never be selected here.


armanbilge (approved):
thanks!

PR #1270, opened 2023-08-01
Fix `test("capture value in closure")`

### LINKED ISSUE #1225: Use of bug in `cfor` macro
The `cfor` macro accidentally generated the desired code by using a bug in beta-reduction. This bug will be patched https://github.com/lampepfl/dotty/pull/16390 and `cfor` will need to be updated.

See https://github.com/dotty-staging/spire/compare/7f630c0209e327bdc782ade2210d8e4b916fddcc...99ea909d086e28e85dbdf8aa78ef0a83bf873405#diff-9cd0f3a9df06a2589472aa8dc960fd2a65a2b65f2f8c257addda1c86ebc31469

froth:
The test broke with the update of scala 3.3.0 (#1225). However the old while loop used in the test as verification did **not** capture the var so always created List(3,3,3). So it tested that the implementation is actually wrong. I changed the test to now use a fixed List.

fixes #1225

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,13 +28,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.13.10, 3.2.2]
+        scala: [2.13.10, 3.3.0]
         java: [temurin@8, temurin@11, temurin@17]
         project: [rootJS, rootJVM, rootNative]
         exclude:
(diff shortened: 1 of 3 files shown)


froth:
This includes an update of the scala 3 version. 

Therefore it supersedes #1249

mio-19:
Is the intended behavior used in real code? If not, is introducing a breaking change an option?

PR #1293, opened 2023-10-24
Update index.md

artemkorsakov:
Syntax `jmh:run` is deprecated.

```
[IJ]jmh:run
[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: Jmh / run
```

May be it is better to use `Jmh / run`?


armanbilge (approved):
Thank you!

PR #1310, opened 2023-12-26
inline algebra typeclass methods

### LINKED ISSUE #1309: RFE: inline methods for spire algebra typeclass instances
I believe it could be an improvement, especially for scala-3, to provide inline definitions of algebra typeclass methods.

To try and illustrate, consider this example.  Here I will define a couple of variations on an `add` function, and also a definition of additive semigroup whose `plus` method is `inline`:

```scala
object defs:
    import algebra.ring.*
    // an in-line function using semigroup typeclass
    inline def add[V](x: V, y: V)(using alg: AdditiveSemigroup[V]): V =
        alg.plus(x, y)
    // a non-inline version of same function
    def add2[V](x: V, y: V)(using alg: AdditiveSemigroup[V]): V =
        alg.plus(x, y)
    // a typeclass with inline 'plus' method
    given AdditiveSemigroup[Double] with
        inline def plus(x: Double, y: Double): Double =
            x + y
    val x = 1.0
```

To compare them, here are two equivalent blocks of code, but one uses the standard Spire algebras and the other uses the inline typeclass above:

```scala
object useSpire:
    import spire.implicits.*
    import defs.*
    val z = add(x, 1.0)

object useInline:
    import defs.*
    import defs.given
    val z = add(x, 1.0)
```

If you compare the resulting bytecode, I believe the typeclass with inline methods gives a more efficient result, using the inline `add` function:

using standard spire algebra:
```
  public static {};
    Code:
       0: new           #2                  // class coulomb/useSpire$
       3: dup
       4: invokespecial #18                 // Method "<init>":()V
       7: putstatic     #20                 // Field MODULE$:Lcoulomb/useSpire$;
      10: getstatic     #25                 // Field spire/implicits$.MODULE$:Lspire/implicits$;
      13: invokevirtual #29                 // Method spire/implicits$.DoubleAlgebra:()Lalgebra/ring/Field;
      16: getstatic     #34                 // Field coulomb/defs$.MODULE$:Lcoulomb/defs$;
      19: invokevirtual #38                 // Method coulomb/defs$.x:()D
      22: invokestatic  #44                 // Method scala/runtime/BoxesRunTime.boxToDouble:(D)Ljava/lang/Double;
      25: dconst_1
      26: invokestatic  #44                 // Method scala/runtime/BoxesRunTime.boxToDouble:(D)Ljava/lang/Double;
      29: invokeinterface #50,  3           // InterfaceMethod algebra/ring/Field.plus:(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
      34: invokestatic  #54                 // Method scala/runtime/BoxesRunTime.unboxToDouble:(Ljava/lang/Object;)D
      37: putstatic     #56                 // Field z:D
      40: return
```
Using algebra with inline method:
```
  public static {};
    Code:
       0: new           #2                  // class coulomb/useInline$
       3: dup
       4: invokespecial #18                 // Method "<init>":()V
       7: putstatic     #20                 // Field MODULE$:Lcoulomb/useInline$;
      10: getstatic     #25                 // Field coulomb/defs$.MODULE$:Lcoulomb/defs$;
      13: invokevirtual #29                 // Method coulomb/defs$.x:()D
      16: dconst_1
      17: dadd
      18: putstatic     #31                 // Field z:D
      21: return
```

If both arguments are literal constants, the inlined improvements are even more pronounced:
```scala
object useInline:
    import defs.*
    import defs.given
    val z = add(1.0, 1.0)  // two constant args
```
scala-3 compiles it down to a single constant value in the byte-code:
```
  public static {};
    Code:
       0: new           #2                  // class coulomb/useInline$
       3: dup
       4: invokespecial #18                 // Method "<init>":()V
       7: putstatic     #20                 // Field MODULE$:Lcoulomb/useInline$;
      10: ldc2_w        #21                 // double 2.0d
      13: putstatic     #24                 // Field z:D
      16: return
```

Lastly, if one uses the non-inline `add2` function, both versions of the typeclass yield the same bytecode, which looks like this:
```scala
object useInline:
    import defs.*
    import defs.given
    val z = add2(x, 1.0)
```

```
  public static {};
    Code:
       0: new           #2                  // class coulomb/useInline$
       3: dup
       4: invokespecial #23                 // Method "<init>":()V
       7: putstatic     #25                 // Field MODULE$:Lcoulomb/useInline$;
      10: getstatic     #30                 // Field coulomb/defs$.MODULE$:Lcoulomb/defs$;
      13: getstatic     #30                 // Field coulomb/defs$.MODULE$:Lcoulomb/defs$;
      16: invokevirtual #34                 // Method coulomb/defs$.x:()D
      19: invokestatic  #40                 // Method scala/runtime/BoxesRunTime.boxToDouble:(D)Ljava/lang/Double;
      22: dconst_1
      23: invokestatic  #40                 // Method scala/runtime/BoxesRunTime.boxToDouble:(D)Ljava/lang/Double;
      26: getstatic     #43                 // Field coulomb/defs$given_AdditiveSemigroup_Double$.MODULE$:Lcoulomb/defs$given_AdditiveSemigroup_Double$;
      29: invokevirtual #47                 // Method coulomb/defs$.add2:(Ljava/lang/Object;Ljava/lang/Object;Lalgebra/ring/AdditiveSemigroup;)Ljava/lang/Object;
      32: invokestatic  #51                 // Method scala/runtime/BoxesRunTime.unboxToDouble:(Ljava/lang/Object;)D
      35: putstatic     #53                 // Field z:D
      38: return
```

erikerlandson:
fix #1309


erikerlandson:
Note, I cannot inline `fromInt` or `fromDouble` because in `Fractional` and `Numeric they inherit multiply from both `Field` and `ConvertibleFromDouble`, and so they _must_ be overridden there.  inline defs are always `final`, so these cannot be inlined at `DoubleIsField`.

Just leaving `fromInt` and `fromDouble` as-is seems reasonable, since fixing this little knot would likely require larger interface changes.

core/src/main/scala-3/spire/std/double.scala
```diff
@@ -0,0 +1,169 @@
...
+  inline def plus(a: Double, b: Double): Double = a + b
+  override inline def pow(a: Double, b: Int): Double = Math.pow(a, b)
+  override inline def times(a: Double, b: Double): Double = a * b
+  inline def zero: Double = 0.0
+
+  override def fromInt(n: Int): Double = n
+
+  override def fromDouble(n: Double): Double = n
+  inline def div(a: Double, b: Double): Double = a / b
+}
```
armanbilge, line 38: Is this the only code that changed? If so, can we avoid duplicating all the code below by leaving it under `scala`?

armanbilge (commented):
Thanks for the PR! In the PR description/comments would you mind sharing a small program whose bytecode would be optimized because of these changes?

core/src/main/scala-3/spire/std/double.scala
erikerlandson: I also am planning on doing the rest of the methods (except conversions)

erikerlandson: Also at least for Int, Long and Float

erikerlandson: I don't think I need to duplicate `DoubleAlgebra` and `DoubleInstances` however, I could move those back under `scala`

erikerlandson: I am being kind of iterative and experimental on this first one, since the full CI sometimes catches things I miss locally.  The others should be faster.

erikerlandson:
> Thanks for the PR! In the PR description/comments would you mind sharing a small program whose bytecode would be optimized because of these changes?

Yes, I can provide an example that invokes at least one method from each instance (Field, NRoot, Trig, etc).  I can exercise all the inlined methods if you want.

armanbilge:
> I can exercise all the inlined methods if you want.

Thanks, just one is sufficient! Essentially a manual test just to demonstrate that this PR actually works, and to demonstrate how to take advantage of it. Since typeclasses are often used in dynamic dispatch situations, I don't think the inlining can work in those cases.

erikerlandson:
> Probably it can only work when the typeclass instance itself is inlined.
That is what my experiments show - if the calling code itself is not `inline`, it will default to the standard non-inlined typeclass calls.

Here is a code example that demonstrates in-lining of methods from "all" the Double typeclasses.  If you compile this in scala-3, you should see that any references to actual typeclass methods have been elided by inlining:

```scala
object inlineTest {
    import spire.implicits.*

    type DoubleTypeClasses = Field[Double]
        & NRoot[Double]
        & Trig[Double]
        & Order[Double]
        & Signed[Double]
        & IsRational[Double]
        & TruncatedDivisionCRing[Double]

    inline def inlinedCalls(x: Double, y: Double)(using alg: DoubleTypeClasses): Double =
        val t1 = alg.plus(x, y)
        val t2 = alg.sqrt(x)
        val t3 = alg.exp(y)
        val t4 = alg.compare(x, y)
        val t5 = alg.abs(x)
        val t6 = alg.tmod(x, y)
        val t7 = alg.floor(y)

        t1 + t2 + t3 + t4 + t5 + t6 + t7

    val x = 1.0
    val y = 2.0

    // with scala-3 inlined typeclasses, the resulting bytecode should
    // show no reference to actual typeclass methods, only raw double
    // operations or calls to 'Math.xxx' methods or other java methods.
    val z = inlinedCalls(x, y)
```

erikerlandson:
@armanbilge  as you can see, this change breaks MIMA.  Before we proceed, is this going to be a problem?

core/src/main/scala-2/spire/std/double.scala
```diff
@@ -0,0 +1,148 @@
...
+import spire.algebra.{Field, IsRational, NRoot, Order, Signed, Trig, TruncatedDivisionCRing}
+import spire.math.Rational
+import spire.util.Opt
+
+import java.lang.Math
+
+trait DoubleIsField extends Field[Double] {
+  override def minus(a: Double, b: Double): Double = a - b
+  def negate(a: Double): Double = -a
+  def one: Double = 1.0
```
armanbilge, line 28: Why did you change this from `val` to `def` on 2.13? It broke bincompat.

armanbilge: Wait, no, it was not this, sorry. Something else changed though on 2.13, let me find it ...

erikerlandson: Oh, probably left over from an earlier experiment and accidentally copied

core/src/main/scala/spire/std/double.scala
```diff
@@ -17,147 +17,14 @@ package spire
...
-  implicit final val DoubleAlgebra: Field[Double]
-    with NRoot[Double]
-    with Trig[Double]
-    with IsRational[Double]
-    with TruncatedDivisionCRing[Double]
-    with Signed[Double]
-    with Order[Double] = new DoubleAlgebra
+  implicit final val DoubleAlgebra
+    : DoubleIsField with DoubleIsNRoot with DoubleIsTrig with DoubleIsReal with DoubleSigned with Serializable =
+    new DoubleAlgebra
```
armanbilge, line 27: The breaking change was this. You can't replace `Field[Double]` with `DoubleIsField`, this breaks bincompat.

core/src/main/scala-2/spire/std/double.scala
erikerlandson: actually, these were `def` before:
https://github.com/typelevel/spire/blob/main/core/src/main/scala/spire/std/double.scala#L28

core/src/main/scala/spire/std/double.scala
armanbilge: To make the change bin-compatibly:

```scala
// keep the old one, but remove `implicit`
final val DoubleAlgebra: Field[Double]
    with NRoot[Double]
    with Trig[Double]
    with IsRational[Double]
    with TruncatedDivisionCRing[Double]
    with Signed[Double]
    with Order[Double] = new DoubleAlgebra

// can't be val b/c bincompat
// hacky but it works
implicit final def DoubleAlgebraBincompat: DoubleAlgebra = DoubleAlgebra.asInstanceOf[DoubleAlgebra]
```

erikerlandson: Unfortunately, if I don't do that, scala-3 won't pick up the `inline` defs and the desired inlining does not happen.

erikerlandson: So the price of getting this improvement is breaking ABI

erikerlandson: It's OK with me if that's something you'd prefer to not do :)

armanbilge: Sorry I'm not sure I understand. You tried my change and it didn't work?

erikerlandson: Oh I missed your proposed fix, I will try that

erikerlandson:
@armanbilge that seems to work  :tada: 

The methods are all inlined.  It seems to be inserting a call to implicit function, although its result is unused in the subsequent bytecode:

```
      20: getstatic     #33                 // Field spire/implicits$.MODULE$:Lspire/implicits$;
      23: invokevirtual #37                 // Method spire/implicits$.ImplicitDoubleAlgebra:()Lspire/std/DoubleAlgebra;
```

Maybe it gets removed in linking?
https://docs.scala-lang.org/scala3/reference/contextual/givens.html#
"The inlining of given instances will not inline/duplicate the implementation of the given, it will just inline the instantiation of that instance. This is used to help dead code elimination of the given instances that are not used after inlining."

I'm a bit puzzled by it, because in my experiments described on #1309, the implicit object never appeared in the byte code at all, but that was an actual object, not an implicit function.

armanbilge:
> Maybe it gets removed in linking?

If by "linking" you mean the JVM optimizes it away at runtime, then maybe? Scala.js and Scala Native have an explicit linking step, but Scala JVM does not.

The reason that static call may be there is to force the companion object to initialize. Not sure.

erikerlandson:
@armanbilge   The four native types are complete.  I think that is where I wanted to get for this.   One minor dangling mystery is that mima check failed for `log` on `NRoot[Int]` and `NRoot[Long]`, I'm guessing because it's exposed somewhere else besides `IntAlgebra`. I removed the inline for that and `fpow` just to see if it passed (it did).

armanbilge:
It should be okay to leave `inline` for `fpow`.

The reason it failed for `log` is because `log` is actually not part of the `NRoot` parent interface. I'm not sure why that method is there. Anyway, when you add `inline`, it takes away the usual method declaration. This hasn't been a problem in the other cases because the method declaration is always inherited.

armanbilge:
> the only question I have left is whether there is a way to get rid of the call to implicit function

I suspect that might be required for semantics, but I'm not sure. Might be a question for the compiler team (I wouldn't be surprised if there was an existing issue on this topic).

erikerlandson:
@armanbilge  I think I have worked out what is going on.  Basically, scala compiles `implicit def ...` differently than `given ...`.  If I declare using `given x: X = ...`, then scala will not include any call to `x()` in its bytecode, but if I use the equivalent `implicit def x: X = ...`, then it will add the `x()` call.

I don't think it's currently feasible to use `given ImplicitDoubleAlgebra ...` and friends with spire, because it also requires the user to `import xxx.{*, given}`, and so doing it breaks the spire code, and presumably any client code currently calling spire.

I could try filing a compiler issue about this, but even if it is feasible to get the compiler to understand when it can elide this call, I doubt it would be a high priority for the compiler devs.

erikerlandson:
@armanbilge  given the above, I think this is the best possible solution for spire, so I'm OK to merge if you are.

I suppose it might be feasible to add some new import that uses `given` for scala-3 users, but that seems like a separate discussion.

armanbilge:
> I could try filing a compiler issue about this, but even if it is feasible to get the compiler to understand when it can elide this call, I doubt it would be a high priority for the compiler devs.

It looks like a bug to me, and specifically a bug with `given`, not with `implicit def`. I don't think it's an optimization issue, it's an issue of correctness / semantics, so I believe it would be high priority. Let's see what they say.
- https://github.com/lampepfl/dotty/issues/19348

erikerlandson:
@armanbilge  the behavior in your reproducer definitely seems like a bug.

erikerlandson:
@armanbilge  so we should consider `@static given`  ?

armanbilge:
I experimented with `@static` today in Cats Effect and it has two issues:
1. I do not see a way to apply it compatibly to existing APIs
2. It's completely broken on Scala.js

https://github.com/typelevel/cats-effect/pull/3934

erikerlandson:
I don't think any `given` declarations can be backward compatible with existing spire, but if we demonstrated they can compile to faster code with scala-3 I could imagine providing some new scala-3 only `given`s to import.

armanbilge:
If we can get `@static` to work, we can just directly apply to `implicit def`s. No need to fuss with `given` for that.

PR #1350, opened 2024-12-10
update sbt-typelevel

samspills:
This is a replacement PR for #1348, just to do the bump in two separate steps and fix a deprecated setting. 
We need this so that the build will keep working in January after a bunch of github actions are removed (upload-artifiact@v3 and download-artifact@v3)

note: the bump to scala 2.13.15 was necessary to resolve some mima issues caused by sbt-typelevel adding the `-Xsource:3` flag. That bump then required bumps to scalajs and native.

### DIFF
--- a/tests/shared/src/test/scala/spire/util/OptSuite.scala
+++ b/tests/shared/src/test/scala/spire/util/OptSuite.scala
@@ -61,8 +61,7 @@ class OptSuite extends munit.FunSuite {
     }
     {
       import spire.std.array._
-      import spire.std.unit._
-      val eq = Eq[Opt[Array[Unit]]]
+      val eq = Eq[Opt[Array[Boolean]]]
       assert(!eq.eqv(Opt(Array.ofDim(0)), Opt.empty))
     }
   }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,12 +15,12 @@ on:
     tags: [v*]
 
 env:
-  PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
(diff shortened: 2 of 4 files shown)


tests/shared/src/test/scala/spire/util/OptSuite.scala
```diff
@@ -61,8 +61,7 @@ class OptSuite extends munit.FunSuite {
     }
     {
       import spire.std.array._
-      import spire.std.unit._
-      val eq = Eq[Opt[Array[Unit]]]
+      val eq = Eq[Opt[Array[Boolean]]]
```
samspills, line 64: This test was failing on JS with 
```
==> X spire.util.OptSuite.Regression test for #896 0.03s java.lang.IllegalArgumentException: null
```

And @armanbilge figured out that it was related to this https://github.com/scala-js/scala-js/issues/5044 + related change in scala.js, and the semantics of `getClass` for unit described here: 
> Scala.js 1.x only: Moreover, for () (unit), the result will be classOf[java.lang.Void] instead of classOf[scala.runtime.BoxedUnit] like the JVM. scala.runtime.BoxedUnit is an implementation detail of Scala on the JVM, which Scala.js does not emulate. Instead, it uses the more sensible java.lang.Void, as Void is the boxed class corresponding to the void primitive type, which is scala.Unit. This means that while java.lang.Void is not instantiable on the JVM, in Scala.js it has a singleton instance, namely (). This also manifests itself in Array[Unit] which is effectively Array[java.lang.Void] at run-time, instead of Array[scala.runtime.BoxedUnit].
-- https://www.scala-js.org/doc/semantics.html#getclass

armanbilge (approved):

PR #1343, opened 2024-08-23
Scala Native 0.5

mio-19:

PR #1363, opened 2025-02-02
Scala native 0.5

### LINKED ISSUE #1342: Scala Native 0.5

arashi01:
- Update to Scala Native 0.5.6
- Update Scala JS to 1.18.2
- Update Scala 3 to current LTS
- Update Scala 2 to 2.13.16
- Update algebra (now that it is within core cats repository)

Build plugins updates:
- update sbt to 1.10.7
- Update sbt-crossproject to 1.3.2
- Update sbt-typelevel to 0.7.7

Resolves #1342

### DIFF
--- a/tests/shared/src/test/scala/spire/algebra/NRootSuite.scala
+++ b/tests/shared/src/test/scala/spire/algebra/NRootSuite.scala
@@ -56,7 +56,7 @@ class NRootSuite extends munit.FunSuite {
     }
   }
 
-  val HighPrecision = new java.math.MathContext(250)
+  val HighPrecision = new java.math.MathContext(if (spire.tests.Platform.isNative) 150 else 250)
 
   val bases = Seq(
     BigDecimal(2),
--- a/tests/shared/src/test/scala/spire/syntax/CforSuite.scala
+++ b/tests/shared/src/test/scala/spire/syntax/CforSuite.scala
@@ -35,6 +35,7 @@ class CforSuite extends munit.FunSuite {
     cfor(0)(_ < 10, _ + 1) { x =>
       cfor(10)(_ < 100, _ + 10) { y =>
         s.add(x + y)
+        ()
       }
     }
     assertEquals(s.toSet, (10 to 99).toSet)
@@ -42,7 +43,7 @@ class CforSuite extends munit.FunSuite {
 
   test("symbol collision cfor") {
     val b = mutable.ArrayBuffer.empty[Int]
-    cfor(0)(_ < 3, _ + 1) { x =>
+    cfor(0)(_ < 3, _ + 1) { _ =>
       cfor(0)(_ < 2, _ + 1) { y =>
         val x = y
         b += x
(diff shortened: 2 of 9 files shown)


armanbilge:
Thanks for the PR! See this other PR that explains why we are struggling to make this upgrade.
- https://github.com/typelevel/spire/pull/1343#discussion_r1730447136

tests/shared/src/test/scala/spire/syntax/CforSuite.scala
```diff
@@ -100,17 +101,21 @@ class CforSuite extends munit.FunSuite {
...
 
   test("capture value in closure") {
-    val b1 = collection.mutable.ArrayBuffer.empty[() => Int]
+    val capturedValues = collection.mutable.ArrayBuffer.empty[Int]
+    val b1 = collection.mutable.ArrayBuffer.empty[() => Unit]
     cfor(0)(_ < 3, _ + 1) { x =>
-      b1 += (() => x)
+      val value = x // Explicitly define. Avoids failures on certain JDKs
+      val closure = () => { capturedValues += value; () }
+      b1 += closure
```
armanbilge, line 109: Are you sure this is related to JDK version? This is actually due to a change in Scala 3.
- https://github.com/typelevel/spire/issues/1225

tests/shared/src/test/scala/spire/syntax/CforSuite.scala
arashi01, line 1: @armanbilge For `CforSuite` I was getting similar failures to #1343 on current versions of Java 8 only. Tested with an old JDK 8 and it wasn't an issue. My guess is there is some change in Java 8's closure expansion? After a bunch of trial and error explicitly declaring a local val seems to fix it.

tests/shared/src/test/scala/spire/algebra/NRootSuite.scala
arashi01, line 1: For `NRootSuite` getting Munit test timeouts running on CI for Native platform. (Running on local workstation is passing). Munit's test timeout setting cannot be changed for scala-native platform. Reduced the precision for Native platform tests. If this is acceptible?

tests/shared/src/test/scala/spire/syntax/CforSuite.scala
arashi01: Interesting. It was definitely passing on older JDK's. Unless I am very confused and was running on scala 2.13!

arashi01: Anyway I believe this implementation is functionally equivalent no? Replaced the while loop with a cfor loop for b2. Can revert that.

armanbilge: To clarify, this is test code, not implementation code. You can consider test code to be representing user code. So if we have to change our test code because the behavior of the implementation changed, that means our users will have to change their code as well. That's not good 😅 when users upgrade, they could be inadvertently introducing bugs.

arashi01:
Closing this as a duplicate of #1343

